PR #174, opened 2015-03-02
Fix bin/jflex scripts to work with mvn package

strubell:
Switched `jflex/bin/jflex` and `jflex/bin/jflex.bat` to look in `target` rather than `lib`, which allows the scripts to work after calling `mvn package` (which puts the jar in `target`).

Note: I didn't test the .bat script as I don't have access to a Windows machine.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -138,8 +138,18 @@
         </plugin>
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
-          <artifactId>maven-clean-plugin</artifactId>
-          <version>2.5</version>
+            <artifactId>maven-clean-plugin</artifactId>
+            <version>2.5</version>
+            <configuration>
+                <filesets>
+                    <fileset>
+                        <directory>lib</directory>
+                        <includes>
+                            <include>**/${project.artifactId}-${project.version}.jar</include>
+                        </includes>
+                    </fileset>
+                </filesets>
+            </configuration>
         </plugin>
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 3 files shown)


lsf37:
Thanks for that idea, it's been annoying for a while that the scripts don't work in the repo version.

Unfortunately your fix will only work on the repo version and not in the release, because the `target` and `build` directories are not included in the release.

What we should probably do instead is to extend the build with a final copy of the jar to `lib` instead. This will also simplify the release scripts, which do this at the moment. (Those scripts are not in the repo yet, so no need to find and fix them).

Would you be interested in trying that? It will be a few days before I can look into it myself.

strubell:
Ah, I thought there might be some deeper issue involved. 

I just pushed some changes 1) reverting the scripts to use the jar in the lib folder and 2) setting the maven jar plugin to output the jar to the lib folder rather than target, which it turns out was really easy!

This setup is working for me, but there could be other things I'm still overlooking.

strubell:
Ok, one thing I overlooked is that mvn clean does not remove the lib jar, but that should be easy enough to fix.

strubell:
Ok, clean should now properly clean the file.

pom.xml
```diff
@@ -164,6 +174,9 @@
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-jar-plugin</artifactId>
           <version>2.4</version>
+          <configuration>
+            <outputDirectory>lib</outputDirectory>
```
regisd: I'm not sure we want to replace the output location of the jar. Most Maven users expect the build jar to be in `/target`

I'd rather have an extra copy step.

lsf37:
Copying files in maven was more interesting than I anticipated..

@strubell, I've transplanted your commit https://github.com/strubell/jflex/commit/7d2b13c0cdd090c3d62a7f96f2e24751bda60e60 on to master (a7df24525ebcb1c851e72e387a3cc9ddc10421a2) and added the file-copy task to both the maven and ant builds, so will close this pull request. Let us know if there are still things that don't work.

Issue #122, opened 2015-02-15
[Bug] suppress unused macro warnings in quiet mode [sf#121]


lsf37:
**Reported by lsf37 on 2012-07-26 05:39 UTC**
If macros are collected in a set of include files, there is a legitimate use case for defining a possibly large number of macros without ever using all of them.

Robert Lieb requests to suppress these warnings in quiet mode. Another solution would be to make certain warnings optional via command line switch/configuration options.

PR #176, opened 2015-03-05
New option —warn-unused

regisd:
New `Options.unused_warning`, which, when set to false, hides warning about unused macros.

For backward compatibility, warnings are enabled by default. On the command line, warnings are disabled in quiet mode `-q` or with `—warn-unused`.

The Maven plugin has a new option `unusedWarning`.

The ant task has a new option `unusedWarning`.

Fix issue #122

The same option could be used for other unused warnings.


lsf37:
Cool!

If the default is `true` for `verbose` mode, we will also need a `--no-warn-unused` option to switch it off on the command line when `verbose` is on (not a problem in ant/maven). Could you add that?

ps: not sure why the test fails and wants to connect to X11, might be able to have a look at this later tonight.

PR #184, opened 2015-04-05
Add a try/finally to close the LineNumberReader to avoid resource leak

regisd:
Add a try/finally to close the LineNumberReader to avoid an unimportant resource leak.

Also reinvent with 2 spaces.


lsf37:
Good idea. Have rebase and merged it into master.

Issue #183, opened 2015-04-02
Warn for probably wrong unicode escapes such as \u0ffff


lsf37:
`\u0ffff` means `\u0fff` followed by the letter `f`. The user probably meant `\U00ffff` instead.

We should add a warning for such occasions (for `\u` and `\U`), which can optionally be suppressed.

See also #179

Issue #179, opened 2015-03-16
classes of >16bit Unicode characters


didier31:
Hi again the jflex team,

I've profited of jflex 1.6 to uncomment classes like [\u116C0-\u116C9] in my lexer.
Talking about this class, my test reveals that jflex generated lexers that this class in particular (but some others too) includes some ponctuation signs characters like  ';' and ':' 

If you need an example, you can refer to the one in https://github.com/jflex-de/jflex/issues/178.

In this example,
it affects IDENTIFIER lexeme in whatever the case modes.
For example, string "myInterger:Integer;" is computed as a whole identifier thanks to  [\u116C0-\u116C9]. This class is part of definition of macro, NUMBER_DECIMAL8.

Best regards
Didier.

lsf37:
Ah well, looks like we might need a 1.6.2 release..

Will need a few days this time before I can take a closer look, am travelling at the moment. Maybe @sarowe has time before.

didier31:
No worry. I'm not hurry. It is just an hobby project of my own.

Happy if I may contribute to make your release more mature.

Have a good journey

Best regards

lsf37:
Thanks, am back now ;-) I've confirmed a smaller example:

```
%%

%public
%class Test

%int
%debug

TAKRI_DIGIT = [\u116C0-\u116C9]
LETTER = [a-zA-Z]

%%

{LETTER} ({LETTER}|{TAKRI_DIGIT})*    {  /* identifier */  return 1; }
[^]                                   {  /* single char */ return 2; }
```
Input: `myInterger:Integer;`

Steps:
```
> jflex -q 179.flex
> javac Test.java 
> java Test input.txt 
match: --myInterger:Integer;--
action [14] { /* identifier */  return 1; }
1
match: --\u000A--
action [15] { /* single char */ return 2; }
2
-1
```

lsf37:
This time the problem is just a syntax error: `\u116C0` doesn't mean code point `116C0`, but instead `\u116C` followed by the character `0`. This means your character class includes `\u115C`, `0-\u116C`, and `9`, which includes the `:` that gets matched in the identifier.

To get code point `116C0`, you can either write `\U0116C0` or `\u{116C0}`:

```
%%

%public
%class Test

%int
%debug

TAKRI_DIGIT = [\U0116C0-\U0116C9]
LETTER = [a-zA-Z]

%%

[...]

PR #188, opened 2015-04-11
Lintifier of JFLex grammar with PMD

regisd:
Hello Gerwin, hello team!

In issue #183, Gerwin suggested to warn users about incorrect usage of Unicode characters.

The easy fix is to change the code and output a warning on stderr.

Grammars can be complex, and I wanted to try a different and more generic approach, based on a _lintifier_, such as Java lint or JavaScript lint. Luckily PMD can be extended for any language :-)

That's what I do in this proof-of-concept. I added a parser for JFlex grammars, and a UnicodeRule that warns you if you use `\u1234x` if `x` is in [A-F0-9].

The good things:
- PMD is easy to integrate with Maven / ant / many continuous integration tools
- PMD has a plugin architecture, this pull request is to discuss `flex-pmd-plugin`
- I'm glad JFLex has it's own grammar, I didn't have to write a tokenizer ;-) I managed to get the list of tokens of the grammar, with `LexScan`.

Where I struggle a bit:
- Ideally the PMD Rules work on the AST. I'm a bit confused… How can I get the AST of JFlex grammar I just tokenized? I got really confused with `LexParse` (I'm really ashamed of asking this question, but I'm sure you'll have a quick answer)
- When I have a token, I'd like to know the starting line, ending line, starting column, ending column (this is for PMD to highlight the error). How do I get those (particularly in multiline scenarios)?
- maven-shade-plugin fails on Travis

Open questions
- Should this be part of jflex, or live in its own directory?
  - Being in the same repository helps with version management


regisd:
If you want to test yourself
* Build the PMD plugin. This should be as easy as `mvn package`
* Install PMD binaries (I haven't done any test or integration with Maven yet)
* In your workspace, run
    pmd.sh pmd -auxclasspath jflex-pmd-plugin/target/jflex-pmd-plugin-1.7.0-SNAPSHOT.jar -R jflex-basic -dir jflex-pmd-plugin/src/test/resources/

This will execute PMD on all files in `jflex-pmd-plugin/src/test/resources/`
There is only one grammar file at the moment there, taken directly from issue #179, which contains on line 9: TAKRI_DIGIT = [\u116C**0**-\u116C**9**]

This should output:

    jflex-pmd-plugin/src/test/resources/UnicodeRule.bad.jflex:8:	\u116C is followed by character 0
    jflex-regisd/jflex-pmd-plugin/src/test/resources/UnicodeRule.bad.jflex:8:	\u116C is followed by character 9

Then of course PMD rules come with a description, and users will know why this warning was thrown.

regisd:
If you use XML output, you even have the column numbers.

		CLASSPATH=/Users/regisd/workspace_jflex/jflex-regisd/jflex-pmd-plugin/target/jflex-pmd-plugin-1.7.0-SNAPSHOT.jar ; /Users/regisd/Downloads/pmd-bin-5.3.0/bin/pmd.sh pmd -auxclasspath $CLASSPATH -R jflex-basic -dir jflex-pmd-plugin/src/test/resources/ -f xml

Outputs:
	
		<?xml version="1.0" encoding="UTF-8"?>
		<pmd version="5.3.0" timestamp="2015-04-11T21:33:50.111">
		  <file name="/Users/regisd/workspace_jflex/jflex-regisd/jflex-pmd-plugin/src/test/resources/UnicodeRule.bad.jflex">
		    <violation beginline="9" endline="9" begincolumn="22" endcolumn="23" rule="MistypedUnicodeCharacter" ruleset="Basic XML" priority="5">
		        \u116C is followed by character 0
		    </violation>
		    <violation beginline="9" endline="9" begincolumn="30" endcolumn="31" rule="MistypedUnicodeCharacter" ruleset="Basic XML" priority="5">
		        \u116C is followed by character 9
		    </violation>
		</file>
		</pmd>
	
As I said previously, I'm not sure how to handle multiline yytext(), though

regisd:
And for some reason, maven-shade-plugin fails on Travis

lsf37:
Hm, this is an interesting one.

On the one hand I feel linting tools should ideally be unnecessary because the compiler should already warn of everything the linting tool can find. Sometimes that's not practical, because the lint analysis is too expensive for a compiler, but that doesn't really apply in our case.

On the other hand, the integration with existing tools and reporting frameworks like PMD is really nice.

So I'm not really sure where I stand on this one. For this specific warning, I think I'd definitely want the standalone version of jflex to have at least the ability to warn about potentially wrong unicode sequences, because they are hard to spot manually and most likely a mistake.

That doesn't mean we can't have the PMD plugin, though, esp if there are other warnings we can think of (initially I thought there would be plenty of things to warn about and added the class [`SemCheck`](https://github.com/jflex-de/jflex/blob/master/jflex/src/main/java/jflex/SemCheck.java), but then not much turned up to implement there. There may be more things on the purely syntactic level, though.

Maybe the use case for the PMD plugin depends mostly on the workflow. You could have all warnings off in the build itself, and generate a readable report with PMD separately. That sounds like something you might want to do in a maven environment. If you're on the command line, you probably want jflex itself to produce the warnings.

lsf37:
To answer some of your questions:

> Ideally the PMD Rules work on the AST. I'm a bit confused… How can I get the AST of JFlex grammar I just tokenized? I got really confused with LexParse (I'm really ashamed of asking this question, but I'm sure you'll have a quick answer)

No need to be ashamed, there is pretty much zero design documentation. I'm amazed I still remember enough to understand the code myself ;-).

There is no full AST for JFlex specs and the parser is mildly horrifying. My only excuse is that I was a student when I wrote it, basically too lazy to write lots of AST classes (These were Java 1.1 days, so these things were pretty verbose. I actually went off to write a tool for generating AST classes a year or so later). It might be a good idea to redesign and refactor the parser, but it's a bit of a bigger job.

If you look at `LexParse.cup`, the toplevel rule is
```
specification ::=  USERCODE
                   /* delimiter is checked in lexer */
                   macros
                   DELIMITER 
                   rules
```
None of them are bound to names, and used in the action that comes after the rule. Instead, the parser relies on (parser-)global state being accumulated in the parse. `USERCODE` is managed in the scanner, `macros` is defined as a list of macros later down, and after parsing each full macro definition, that macro is added to a global list `macros` of class `Macros`, which is a member field of the parser action class. Similarly for the regexp rules with actions: the grammar defines a list of rules `rules` and manages lexical state information (`<YYINITIAL>` etc) for rules that are already parsed. The grammar rule `rule` parses a single regexp rule with action and stores it in a global `RegExps` container.

So in a way, the AST for JFlex is the user code stored in the scanner, the `Macros` container stored in the parser action object, and the `RegExps` container also stored int the parser action object.

The regexps themselves do have an AST - a pretty type-unsafe one, consisting basically of the classes `RegExp1` and `RegExp2`, where both use class `sym` to identify what kind of node they are and `RegExp1` has one terminal value (like a character or string) or one content regexp (like `*`), and `RegExp2` has two regexp child node (like `|`).

> When I have a token, I'd like to know the starting line, ending line, starting column, ending column (this is for PMD to highlight the error). How do I get those (particularly in multiline scenarios)?

I think you have these in a later commit already: start line and column should be in the lexer. If you are lucky, there is a function somewhere in the lexer that computes end line and column from start and `yytext()`, but it's possible that we've never needed that info before, because jflex usually just prints the first line for warnings and errors.

> maven-shade-plugin fails on Travis

Not sure about that one either. I haven't had time yet to try the plugin out myself. Will report back if it fails for me too.

> Should this be part of jflex, or live in its own directory?
Being in the same repository helps with version management

My main concern would be to avoid a runtime dependency of main JFlex on pmd (a build dependency is fine). Can we rely on the class loader to not touch/fail on the pmd classes if they are not used?

regisd:
I give up on this idea.

PR #209, opened 2017-02-13
Add support for Unicode 9.0

sarowe:


sarowe:
I noticed two things when I (accidentally) generated the Unicode data files with Java 8:

* Some of the script extension property codepoint ranges were differently ordered, but they shouldn't be, since they should be strictly sorted.
* Some of the script extension property ranges were *previously* improperly sorted.

After some investigation I discovered that jflex-unicode-maven-plugin's UnicodeVersion wasn't properly merging or sorting added ranges.  The first commit on this pull request also fixes this problem; as a result, there are changes in most Unicode versions' data files, including in other properties than script extension.

PR #212, opened 2017-04-09
fix typo

iorixxx:
Typo: unkown => unknown


lsf37:
Hi @iorixxx, thanks for fixing that typo!

I'll investigate the CI failure (probably benign) and then merge after that.

lsf37:
Ok, I figured out what's breaking the test: you fixed the typo also in the test input (testsuite/testcases/src/test/cases/java/java-2.input). This changes the test output not only in content, but also in position information, which is what it is complaining about.

I'll merge and regenerate a known-good test output.

PR #213, opened 2017-04-10
Marked plugin as thread safe

### LINKED ISSUE #211: Maven warns that plugin not thread safe
When I execute mvn package -T8, the following warning will be displayed:

[WARNING] *****************************************************************
[WARNING] * Your build is requesting parallel execution, but project      *
[WARNING] * contains the following plugin(s) that are not marked as       *
[WARNING] * @threadSafe to support parallel building.                     *
[WARNING] * While this /may/ work fine, please look for plugin updates    *
[WARNING] * and/or request plugins be made thread-safe.                   *
[WARNING] * If reporting an issue, report it against the plugin in        *
[WARNING] * question, not against maven-core                              *
[WARNING] *****************************************************************
[WARNING] The following plugins are not marked @threadSafe in XXX:
[WARNING] de.jflex:jflex-maven-plugin:1.6.1
[WARNING] *****************************************************************

dheid:
Updated to maven-plugin-api 3.0.5 and introduced Mojo annotations. Marked JFlexMojo as thread-safe. Closes #211

### DIFF
--- a/jflex-maven-plugin/pom.xml
+++ b/jflex-maven-plugin/pom.xml
@@ -44,7 +44,7 @@
     <url>https://github.com/jflex-de/jflex/issues</url>
   </issueManagement>
   <prerequisites>
-    <maven>3.0.1</maven>
+    <maven>${maven.version}</maven>
   </prerequisites>
   <inceptionYear>2007</inceptionYear>
   <developers>
@@ -90,6 +90,7 @@
   </licenses>
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <maven.version>3.0.5</maven.version>
   </properties>
   <build>
     <plugins>
@@ -102,10 +103,6 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
         <version>3.1</version>
-        <configuration>
-          <source>${jflex.jdk.version}</source>
-          <target>${jflex.jdk.version}</target>
-        </configuration>
       </plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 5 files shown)


sarowe:
This looks great, thanks David!  But as Gerwin mentioned over on #211, there are known non-thread-safe things that need to be dealt with, e.g. converting static options to instance members, before we can declare the plugin thread-safe.

dheid:
Is it possible to add at least the annotations and mark the plugin as non thread-safe until the conversion will be done? I think, the annotations are a useful improvement in general.

lsf37:
I think the annotations are a good idea, happy to have them, if we mark it as non thread-safe for now. @sarowe do you agree?

sarowe:
> I think the annotations are a good idea, happy to have them, if we mark it as non thread-safe for now. @sarowe do you agree?

+1

dheid:
Nice!!! Thank you!

lsf37:
Now merged and marked as not thread safe (for now) as discussed in 104b2197468039d132578b183ed44ed0d8ede503

Thanks for the contribution!

jflex-maven-plugin/pom.xml
```diff
@@ -44,7 +44,7 @@
     <url>https://github.com/jflex-de/jflex/issues</url>
   </issueManagement>
   <prerequisites>
-    <maven>3.0.1</maven>
+    <maven>${maven.version}</maven>
```
regisd: We need a very recent version of Maven to compile JFlex due to a bug in maven-shade-plugin.

Are you sure we actually need the same to execute the plugin? It sounds to me that 3.0 should be enough.

regisd:
Thanks for updating the plugin to use Java annotations rather than invalid javadoc tags.

jflex-maven-plugin/pom.xml
dheid: Maven version 3 should really be enough. The annotation were introduced  back then.

dheid:
You are welcome!

PR #218, opened 2017-10-31
Fix simple spell check issues

kinow:
Hi, while fixing some spelling mistakes in the lucene-solr project, I noticed that some were originating from jflex. So removed the corrections from the Java files in lucene-solr, but thought it would be worth submitting a pull request here too just in case?

This pull request fixes some minor spelling mistakes found in Java files.

Before the changes, `run-tests` reports 1 failure. After the changes, it reports the same test failing. Had a look but without knowing much about jflex, it wasn't clear how the test could be fixed, sorry.

Hope it helps
Bruno

### DIFF
--- a/jflex/examples/interpreter/SymtabEntry.java
+++ b/jflex/examples/interpreter/SymtabEntry.java
@@ -12,7 +12,7 @@
  * Symbol table entry for names, there are subclasses for
  * variables and functions.
  * 
- * Defines constants UNKNOWN, VAR und FUN as kinds of
+ * Defines constants UNKNOWN, VAR and FUN as kinds of
  * symbol table entries.
  */ 
 class SymtabEntry {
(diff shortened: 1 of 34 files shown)


regisd:
Thanks kinow. I don't see how you can break anything. But let's close and reopen this pull request to trigger Travis with the fixed configuration.

regisd:
Travis, is code compiling and test passing?

kinow:
I'll have a look at how Travis is running it later. On my Ubuntu LTS + JDK8, I had one annoying test always failing. But this test was failing after I checked out a fresh copy of the master branch too. Probably something is missing in my environment, or I missed some configuration.

Thanks for triggering another build.

lsf37:
This looks good now, will merge. Thanks for fixing all of these, a bit embarrassing that I didn't do this ages ago already..

kinow:
These typos are really trivial. And the code - which is the most important part - seems to be working great (not much of a user, but I use Solr/Lucene that I believe use jflex).

Thanks for merging it and for jflex :-)

Issue #222, opened 2017-11-01
jflex has error-prone fall-through


regisd:
jflex/jflex-unicode-maven-plugin/target/generated-sources/jflex/jflex/ArchaicLineBreakScanner.java:702: error: [FallThrough] Execution may fall through from the previous case; add a `// fall through` comment before this line if it was deliberate
```
          case 11: break;
          ^
```

see http://errorprone.info/bugpattern/FallThrough

lsf37:
The fall-through violations seem to be in generated code -- we could automatically add the `// fall through` comment in the generator to indicate that this is on purpose.

lsf37:
Right, that's actually what that specific fall through is for: to guard for the cases where there is no `return` or `throw` and the scan just continues. Do we get warnings for fall through comments that are spurious? If not, we could add another line with something like
`// allowed to fall through if action does not return or throw`

wrprice:
Curious... why use fall through to a new case instead of adding an explicit `break;` ?

regisd:
`// fall through` is the exact opposite of `break`. It means the case falls to the following case statement, and that the developer has not forgotten the `break`.

wrprice:
I understand, but I noticed quite a few cases that fall-through to cases that _only_ `break;` and do nothing else.  I was just about to edit my post because the answer is probably because you don't know whether the action code included a return or break, and without examining it the safest thing is to fall-through to a separate case that breaks.  Blindly inserting a break after an action could result in an unreachable statement otherwise.

Sorry for the noise.

regisd:
There are still violations 
https://travis-ci.org/jflex-de/jflex/jobs/431900782

regisd:
Work around: Use javac option `-Xep:FallThrough:WARN`

regisd:
I also believe that JFlex 1.7.0 has fall-through violations that 1.6.1 didn't have.

PR #223, opened 2017-11-01
Add google error-prone in compile step.

regisd:
Note that FallThrough violations have been reduced to warnings.

issue #222

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -126,6 +126,36 @@
   </dependencyManagement>
     
   <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>3.3</version>
+        <configuration>
+          <compilerId>javac-with-errorprone</compilerId>
+          <forceJavacCompilerUse>true</forceJavacCompilerUse>
+          <compilerArgs>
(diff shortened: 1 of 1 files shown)


lsf37:
The test seems to have failed on something strange:
`The command "mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V" failed and exited with 1`

Any idea where that is from?

regisd:
That's from:
> [ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.3:compile (default-compile) on project jflex-unicode-maven-plugin: Execution default-compile of goal org.apache.maven.plugins:maven-compiler-plugin:3.3:compile failed: An API incompatibility was encountered while executing org.apache.maven.plugins:maven-compiler-plugin:3.3:compile: java.lang.UnsupportedClassVersionError: javax/tools/DiagnosticListener : Unsupported major.minor version 52.0

The error-prone plugin is compiled with Java 8, and Travis uses openjdk7. See #225

lsf37:
Right, let me ping Travis again to see if it all works on jdk8. Could we run the plugin conditionally when jdk8 is available?

regisd:
So, indeed

- 155.1 on openjdk7 fails
- 155.2 on openjdk8 passes

PR #219, opened 2017-10-31
Fix FindBugs issues

kinow:
Found a few Maps being iterated twice while preparing another pull request. Then triggered a `mvn clean findbugs:findbugs findbugs:gui`, and fixed a few ones. No regressions found running `./run-test`

PR #224, opened 2017-11-01
Fix FindBugs issues

regisd:
- Performance: Inefficient map iterator
  Improve loops on maps by reading the entry, rather than the key and then the value
- Use AutoClosable
- Performance: Use Integer.valueOf() instead of creating new Integer objects

```
mvn clean
mvn -N install
mvn package findbugs:findbugs
```


regisd:
#219 was submitted in commit/cfbb7094569b524631e47ca54e9f5268ede223d5 but I rewrote the history in master…

PR #226, opened 2017-11-01
Format java code automatically

regisd:
Use google-java-format to format code.

This is invoked automatically in the compile step thanks to the fmt-maven-plugin.


regisd:
For consistency, I suggest we use the google-java-format tool.

I know this will taint blame history, but I think it's a good investment for the future.

The google style is very reasonnable, but I'm happy to replace by another plugin if you prefer a different style (by design, this tool has no option).

lsf37:
I'm in principle happy with this, but we shouldn't apply it to upstream files like java_cup.*

@sarowe do you have any strong feelings on this?

regisd:
That's a good point.

the plugin has a `sourceDirectory` option that could be used to include only jflex sources.

On the other hand, do we need the cup source code in the jflex repo? I'm wondering if we couldn't fetch the cup source from https://versioncontrolseidl.in.tum.de/parsergenerators/cup.git in the Maven lifecycle.

sarowe:
> @sarowe do you have any strong feelings on this?

I'm fine with it.

lsf37:
It's only the cup runtime (as opposed to the generator itself) that we need, but for that one I would prefer to keep a local copy in the repo.

The main reason is that the upstream location of cup has changed multiple times during the life cycle of jflex and if we don't keep it locally, it's going to be extremely hard to go back in history and produce a working version.

`sourceDirectory` sounds like a good option.

lsf37:
Just looked through the new commits: does the pom change mean that the cup runtime is in a separate `.jar`? If yes, that might mess with the ability to run jflex by double clicking the `jar` (or not, but we should, I guess).

regisd:
I couldn't find a way with `sourceDirectory`. Externalizing cup in its own package sound cleaner, but clearly broke Jflex, because the test doesn't pass anymore:
```none
Compile failed; see the compiler error output for details.
warning: [options] bootstrap class path not set in conjunction with -source 1.7
warning: Supported source version 'RELEASE_6' from annotation processor 'org.eclipse.sisu.scanners.index.SisuI$
jflex/testsuite/testcases/src/test/cases/cupsym/Cupsym.java:10: error: package j$
public class Cupsym implements java_cup.runtime.Scanner {
                                               ^
jflex/testsuite/testcases/src/test/cases/cupsym/Cupsym.java:464: error: package $
  public java_cup.runtime.Symbol next_token() throws java.io.IOException {
```
I'm looking where to add the dep to make it work again.

regisd:
I have been hitting this error many times
>[ERROR] Failed to execute goal org.apache.maven.plugins:maven-shade-plugin:3.1.0:shade (default) on project jflex: Error creating shaded jar: /home/travis/build/jflex-de/jflex/cup_runtime/target/classes (Is a directory) -> [Help 1]

https://travis-ci.org/jflex-de/jflex/jobs/297031421

I just managed to reproduce locally with Maven 3.5.0

I don't have this issue with Maven 3.2.2 or Maven 3.5.2.

So now, I have to find how to work around a Maven bug…

PR #230, opened 2017-11-04
Cleanup of the POM

regisd:
Remove reporting section that we don't use.

Declare all dependencies in dependencyManagement.


lsf37:
I've rebased over current master. Please have a look if I messed up anything -- happy for this to be merged from my side.

regisd:
Please hold. I think I broke the site generation.
At the same time, do we care about mvn site? They look ugly and I don't think we even publish them.
I liked the different reports they had, but I now believe these reports should actually be part of the testing suite.

lsf37:
No worries. I'm not sure wether we published the site or not. I think we used to, but haven't for a while. Happy not to.

PR #231, opened 2017-11-04
Reformat java code with google-java-format

regisd:
Recompile locally to benefit from google-java-format added in #226


regisd:
Since you were ok in #226 I'll submit this

PR #235, opened 2017-11-04
Add fall-through statements in Emitter

regisd:
Add `// fall through` statements in switch/case.

I think this fixes the error-prone FallThrough violations (#222).

However, because of bootstrapping, the test can be re-enabled only in jflex 1.7.1 or later.

### DIFF
--- a/jflex/src/main/java/jflex/Emitter.java
+++ b/jflex/src/main/java/jflex/Emitter.java
@@ -920,10 +920,10 @@ public final class Emitter {
       println("        zzCh = Character.codePointAt(zzBufferL, zzCurrentPosL, zzMarkedPosL);");
       println("        zzCharCount = Character.charCount(zzCh);");
       println("        switch (zzCh) {");
-      println("        case '\\u000B':");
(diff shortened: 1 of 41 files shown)


jflex/src/main/java/jflex/Emitter.java
```diff
@@ -921,10 +921,10 @@ private void emitLexFunctHeader() {
...
       println("        zzCharCount = Character.charCount(zzCh);");
       println("        switch (zzCh) {");
-      println("        case '\\u000B':");
-      println("        case '\\u000C':");
-      println("        case '\\u0085':");
-      println("        case '\\u2028':");
+      println("        case '\\u000B':  // fall though");
+      println("        case '\\u000C':  // fall though");
+      println("        case '\\u0085':  // fall though");
+      println("        case '\\u2028':  // fall though");
```
wrprice: @regisd - I know this was merged a long time ago, but "through" is consistently misspelled, FYI.

regisd: Indeed, thanks!

PR #236, opened 2017-11-04
Integrate Maven plugin documentation in manual

regisd:
Integrate Maven plugin documentation in manual rather than in its own site (which hasn't been published for a while).


lsf37:
This is a good idea!

There are a few small issues to work out first:
 - integration into html and latex/pdf generation
 - syntax highlighting: the other bits of the manual don't have any, mostly because we don't have it for jflex and it looks inconsistent if some sections do and some don't. I'd like to switch it off for now.
 - section hierarchy: the plugin section is one level too high, it should come under "installing"
 - make sure it looks good in all 3 output versions

I've started on these locally, should be able to finish them later today and push.

docs/Makefile
```diff
@@ -12,12 +12,14 @@ build: realclean all
...
-              md/encodings.md md/performance.md md/porting-and-parsers.md md/end.md
+md/manual.md: md/head.md md/intro.md md/installing.md \
+              md/maven-plugin.md md/ant-task.md md/example.md md/lex-specs.md \
+              md/encodings.md md/performance.md md/porting-and-parsers.md\
+              md/end.md
 	cat $^ > $@
 
 manual.html: manual.mdx
-	pandoc -S --css manual.css --filter pandoc-citeproc -f markdown $< -o $@
+	pandoc --css manual.css --filter pandoc-citeproc -f markdown+smart $< -o $@
```
lsf37: The most recent pandoc install I'm getting with macports doesn't seem to support the `+smart` target. Is it essential? What does it do?

lsf37: Just saw in the man page that this is the same as `-S` or `--smart`. Would it be Ok to leave in the command line option or does that create a conflict with newer versions?

lsf37:
Ok, I've added a bunch of commits to address these. Please have a look and feel free to merge if you're happy with them.

docs/Makefile
```diff
@@ -19,13 +19,13 @@ md/manual.md: md/head.md md/intro.md md/installing.md \
 	cat $^ > $@
 
 manual.html: manual.mdx
-	pandoc --css manual.css --filter pandoc-citeproc -f markdown+smart $< -o $@
+	pandoc --smart --css manual.css --filter pandoc-citeproc -f markdown $< -o $@
 #	pandoc --filter pandoc-citeproc -f markdown --toc --toc-depth=2 $< -o $@
```
regisd: I changed this because my version of pandoc fails with
> --smart/-S has been removed.  Use +smart or -smart extension instead.

regisd: I'll merge, because fixing this is a bit unrrelated, and the rest LGTM

PR #238, opened 2017-11-04
Update address of central Maven repo in ant script

regisd:
The central.maven.org has been replaced by repo.maven.apache.org


lsf37:
Thanks for that!

PR #240, opened 2017-11-04
Make java code Doclint compliant

### LINKED ISSUE #182: Make generated code pass doclint
While I have yet to find anyone who actually likes it, doclint is enabled by default when compiling with JDK 8.  It would be nice if generated lexers passed doclint inspection without errors or warnings.

regisd:
I added javadoc with
```
mvn javadoc:fix
```
Hence, I'm not saying I'm adding test, the javadoc has been generated automatically and then I improved the javadoc only in a couple places.

Fixes #182 With this change, all the tests compile without doclint violation. I'm not sure what's wrong with the generated code.


lsf37:
Ok, let's do that. I'm not sure about basically empty javadoc, but then again we can keep improving it over time.

Will merge. (And thanks!)

PR #244, opened 2017-11-06
Let Travis fail if a test fail

### LINKED ISSUE #242: run-tests can exit as success, even if some tests fail
In pull request #241 , I broke the Makefile-powered tests, and yet, Travis sees tests passing:

```none
Writing code to "Lexer.java"
java -jar ../../lib/java-cup-11b.jar < ycalc.cup
Error: Unable to access jarfile ../../lib/java-cup-11b.jar
make: *** [parser.java] Error 1
./run-tests: line 37: cd: interpreter: No such file or directory
./run-tests: line 38: cd: java: No such file or directory
./run-tests: line 39: cd: simple-maven: No such file or directory
./run-tests: line 40: cd: standalone-maven: No such file or directory
./run-tests: line 41: cd: zero-reader: No such file or directory
./run-tests: line 44: cd: jflex: No such file or directory
The command "./run-tests" exited with 0.
Done. Your build exited with 0.
```

https://travis-ci.org/jflex-de/jflex/jobs/297505025

regisd:
`set -e`: Exit immediately if a command exits with a non-zero status.

However
> the shell does not exit if the command that fails is part of the command list immediately following a while or until keyword, part of the test in an if statement, **part of any command executed in a && or ||** list except the command following the final && or ||, any command in a pipeline but the last, or if the command’s return status is being inverted with !. 

https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html

Hence ending the execution of the test with `&& cd ..` prevents the exit of the script in case of failure.

Change to parent directory on its own line.

Fix #242

### DIFF
--- a/jflex/examples/cup/Makefile
+++ b/jflex/examples/cup/Makefile
@@ -1,8 +1,7 @@
 JAVA=java
 JAVAC=javac
 # Root of the project
-# regis Wrong on purpose, I want to see Travis fail
-ROOT=../..
+ROOT=../../..
 JFLEX=$(ROOT)/jflex/bin/jflex
 CUPJAR=$(ROOT)/cup/cup/java-cup-11b.jar
 CUP=$(JAVA) -jar $(CUPJAR) <
--- a/jflex/examples/cup/Makefile
+++ b/jflex/examples/cup/Makefile
@@ -2,8 +2,8 @@ JAVA=java
 JAVAC=javac
 # Root of the project
 ROOT=../../..
-JFLEX=$ROOT/jflex/bin/jflex
-CUPJAR=$ROOT/cup/cup/java-cup-11b.jar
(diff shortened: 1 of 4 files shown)


lsf37:
Following on from the other discussion, the problem is that the `&&` chain propagates the exit code correctly, but `set -e` ignores it. That means though, that this change only observes failure in the last `make` in the chain on each line. If the directory is wrong (as it was for the ant build), or if something before in the `make clean` goes wrong, we won't see it showing up in Travis. 

If I'm reading `set -e` right, replacing `&&` with just `;` should do the trick, because it's not a `&&` chain any more. I.e. my whole attempt of using `&&` for propagating failure was just wrong unless you use it with the `|| exit 1` pattern. I'll experiment with `;` and see if that is a bit nicer and catches all of them.

PR #248, opened 2017-11-07
Replace the cup ant task by a Maven plugin.

regisd:
* Add a new **de.flex:cup-maven-plugin** with goal **generate**
 - For now, the plugin only suppots options **parser, symbols, symbolInterface**
 - Add a simple project used for integration testing
* Use **de.flex:cup-maven-plugin:generate** in **jflex** project instead of ant task

The benefit is that
- we are more consitent in our use of Maven
- the compilation doesn't start a thread to execute ant


cup-maven-plugin/src/main/java/de/jflex/plugin/cup/GenerateMojo.java
```diff
@@ -0,0 +1,196 @@
...
+
+  /** Constant {@code src/main/cup}. */
+  private static final String DEFAULT_SRC_DIRECTORY = "src/main/cup";
+
+  /** In a CUP definition, the Java package is introduce by the {@code package} keyword. */
+  private static final String PACKAGE_DEFINITION = "package";
+
+  private static final String DEFAULT_JAVA_PACKAGE = "";
+
+  // Seriously – by default CUP generates lower case class names.
```
lsf37: I understand the exasperation, but I'd still remove the 'Seriously' ;-)

cup-maven-plugin/src/main/java/de/jflex/plugin/cup/GenerateMojo.java
```diff
@@ -0,0 +1,196 @@
...
+  String parserName;
+
+  @Parameter(property = "project", required = true, readonly = true)
+  MavenProject mavenProject;
+
+  private CliCupInvoker cupInvoker;
+  private final Logger log;
+
+  /** Whether to force generation of parser and symbols. */
+  // private boolean force;
```
lsf37: Should be remove this one or is it a TODO for later?

cup-maven-plugin/src/test/java/de/jflex/plugin/cup/GenerateMojoTest.java
```diff
@@ -0,0 +1,102 @@
...
+import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+
+/** Tests for {@link GenerateMojo}. */
+public class GenerateMojoTest {
+
+  private static final String TEST_TARGET_GENERATED_CUP_DIRECTORY =
+      "/tmp/target/generated-sources/cup";
```
lsf37: Does this work on Windows platforms? (Not really that important if it doesn't since is for development only, but if there's a platform independent API for tmp files, it would be slightly preferable)

cup-maven-plugin/src/main/java/de/jflex/plugin/cup/CliCupInvoker.java
```diff
@@ -0,0 +1,100 @@
...
+   */
+  void invoke(
+      String javaPackage,
+      File outputDirectory,
+      String parserClassName,
+      String symClassName,
+      boolean symbolInterface,
+      String cupFileName)
+      throws Exception {
+    // Seriously? cup doesn't have a better API than calling main like on cli!
```
lsf37: It'd be nice to phrase that more neutrally ;-) E.g. `CUP only has the command line as API, so we have to interface there.`

CUP is very old, like JFlex, and it hasn't seen as much active maintenance, so there are a few crufty things that one wouldn't write that way any more these days.

regisd: Yes, true. I was between surprise and exasperation. Rephrased.

cup-maven-plugin/src/main/java/de/jflex/plugin/cup/GenerateMojo.java
regisd: Done.

cup-maven-plugin/src/test/java/de/jflex/plugin/cup/GenerateMojoTest.java
regisd: Done.
Good point. I have no idea.Guava has a handy `Files.createTempDir()`.

cup-maven-plugin/src/main/java/de/jflex/plugin/cup/CliCupInvoker.java
regisd: And please use **squash** for merging :-)

lsf37: Yes :-)

PR #259, opened 2017-11-11
Commons cli

regisd:
Use apache commons-cli rather than hand made parser.

Also, simplified the configuration of apache-shade-plugin. The uber jar is now called `jflex-full-{VERSION}.jar`

### DIFF
--- a/jflex/pom.xml
+++ b/jflex/pom.xml
@@ -55,6 +55,10 @@
   </developers>
   <dependencies>
     <dependency>
+      <groupId>commons-cli</groupId>
+      <artifactId>commons-cli</artifactId>
+    </dependency>
+    <dependency>
       <groupId>de.jflex</groupId>
       <artifactId>cup_runtime</artifactId>
       <version>${cup.version}</version>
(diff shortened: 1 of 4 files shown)


lsf37:
Hm. I can see what you are trying to do, but I must say I'm not a big fan of this one, because it introduces an unnecessary runtime (generator time) dependency, it almost does the parsing logic twice (we get two chains, one with `addOption`, one with `hasOption`), and it actually changes the CLI. The single `-longoption` form is unusual on Unixes, but it's been there for almost 20 years and it is what javac uses. There is likely to be a whole bunch of Makefiles out there that use it, and we shouldn't break them if it is avoidable.

Not sure about the `throws` statement in `main`: `main` should be the CLI and not be able to throw an exception, because that means CLI users might be able to see an exception message, which should be excluded by type.

regisd:
Thanks for having a look Gerwin.

- Yes, I changed a couple of "long" options, but that was mostly personal taste (`--nobak` doesn't look good to me). I can revert most of them (except a few, like quiet which has two shorts and one long)
- I don't really see where the parsing is done twice. It's done once, and then the set of values is used to set the jflex Options static fields
- Concerning the `Exception`s thrown in main, I agree it's not very nice, but I wanted to leave the API where `main()` is called by another app. Or maybe I got confused because the API was `genereate(argv)` before. So I think I can also align this with another `generate(real,param,values)` and let `main()` catch everything.

Now, the real question is wether the 50kB of commons-cli is woth it. I don't really think we should parse cli, but you like parsers, so…

PR #262, opened 2017-11-12
Rename uberjar to jflex-full-VERSION.jar

### LINKED ISSUE #261: jflex-testsuite-maven-plugin tests fail if Maven repository location is changed
With `~/.m2/settings.xml`, it is possible to change the path of the local repository.

This causes the jflex-testsuite to fail, with ClassNotFoundException on  jflex.* java_cup.*

regisd:
By default the maven shade plugin:

1. takes jflex.jar
2. shades it
3. builds a shaded jar, overwrites jflex.jar and renames the original original-jfex.jar

I think this is not good for the following reasons:

- The maven-shade-plugin seems to be confused in incremental builds (and warns that java_cup is already part of the original jflex)
- Goes against the usual naming convention. In Maven world, a jar is a lib, not an uberjar.

With this change:

- Rename the uber jar flex-full-${VERSION}.jar
- Update scritps
- Update jflex-testsuite-maven-plugin to get the uberjar by configuration rather than hardcoding a ~/.m2/repository path (fixes #261)
- Also Update jflex-testsuite-maven-plugin Mojo to use annotations rather than at-clauses.


regisd:
@lsf37 Is the impact of renaming the jar acceptable? In a next step, I'd like to split out the GUI (and maybe the ant task as well).

PR #277, opened 2018-07-30
Fixed discrepancy in documentation

SJrX:
I'm not 100% sure of this fix, I was just reading and skimming the documentation, and second line in the change referenced the "next backslash", I didn't understand why the rule would work that way. Looking at the example on line 98 however makes sense, so I believe these should be the same.


lsf37:
You are completely right. I must have made a mistake when I copied that line manually back then.

lsf37:
I'm not sure what it going on with Travis at the moment, but the build error is completely unrelated to this pull request, so I'm going to merge anyway.

Issue #281, opened 2018-08-28
Build failure: UnsupportedClassVersionError when compiling byaccj


regisd:
https://travis-ci.org/jflex-de/jflex/jobs/410185220
```
javac -cp .:../../../cup/cup/java-cup-11b.jar Main.java Lexer.java parser.java
Picked up _JAVA_OPTIONS: -Xmx2048m -Xms512m
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
java -cp .:../../../cup/cup/java-cup-11b.jar Main test.txt > output.txt
Picked up _JAVA_OPTIONS: -Xmx2048m -Xms512m
Error: A JNI error has occurred, please check your installation and try again
Exception in thread "main" java.lang.UnsupportedClassVersionError: Main has been compiled by a more recent version of the Java Runtime (class file version 53.0), this version of the Java Runtime only recognizes class file versions up to 52.0
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:763)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:467)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:73)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:368)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:362)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:361)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:335)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at sun.launcher.LauncherHelper.checkAndLoadMain(LauncherHelper.java:495)
make: *** [output.txt] Error 1
```

regisd:
I guess `javac` is version 7 when we run 

    javac -cp .:../../../cup/cup/java-cup-11b.jar Main.java Lexer.java parser.java
    java -cp .:../../../cup/cup/java-cup-11b.jar Main test.txt > output.txt

If so, that would be a bug in Travis *openjdk7* environment.

regisd:
This gives us
```
openjdk version "1.8.0_141"
javac 9.0.1
```
[...]

PR #282, opened 2018-08-29
Print version of java and javac before running tests

regisd:
Investigate issue #281


regisd:
Obviously not going to fix the tests, but this outputs the detailed version which is useful

PR #289, opened 2018-08-29
Move check java-format in its own test suite.

regisd:


regisd:
Can't find a way to do this elegantly before building everything

Issue #164, opened 2015-02-15
Input/Output file encoding setting [sf#29]


lsf37:
**Reported by *anonymous on 2013-04-15 04:54 UTC**
It seems that current implementation only accepts system default encoding settings for input (.flex) / output (.java) file.

I think both command line argument and ANT script settings should have encoding options.

Suggested command line example:

java JFlex.Main -encoding utf-8 CodeTest.flex

Suggested ANT script entry example:

&lt;jflex file="${src.dir}/CodeTest.flex" destdir="${build.generated.sources.dir}/flex" encoding="utf-8" /&gt;

lsf37:
Addressing review comments

lsf37:
Now also does output file encoding in adcbeee537

PR #301, opened 2018-09-02
Use UTF-8 encoding in example projects

regisd:
Fixes [WARNING] Using platform encoding (UTF-8 actually) to copy filtered resources, i.e. build is platform dependent!

See #164


lsf37:
Looks good!

PR #308, opened 2018-09-16
Encoding

lsf37:
Adding an encoding option for reading lexer specs. Addresses issue #164.


jflex-maven-plugin/src/main/java/de/jflex/plugin/maven/JFlexMojo.java
```diff
@@ -221,6 +228,14 @@ private void parseLexFile(File lexFile) throws MojoFailureException, MojoExecuti
       throw new MojoExecutionException("Illegal generation method: " + generationMethod);
     }
 
+    if (!encodingName.equals("")) {
```
regisd: nit: Prefet Objects.equal()

jflex/src/main/java/jflex/Options.java
```diff
@@ -46,6 +47,8 @@
    * If true, dot (.) metachar matches [^\n] instead of [^\r\n\u000B\u000C\u0085\u2028\u2029]|"\r\n"
    */
   public static boolean legacy_dot;
+  /** The encoding to use for input files. */
```
regisd: What about the output?

jflex/src/main/java/jflex/Options.java
```diff
@@ -88,6 +91,16 @@ public static void setDir(File d) {
     directory = d;
   }
 
+  /** Set encoding for input files, and check availability of encoding on this JVM. */
```
regisd: nit: s/Set/Sets

jflex/src/main/java/jflex/Options.java
```diff
@@ -88,6 +91,16 @@ public static void setDir(File d) {
...
+  public static void setEncoding(String encodingName) {
+    if (Charset.isSupported(encodingName)) {
+      encoding = Charset.forName(encodingName);
+    } else {
+      Out.error(ErrorMessages.CHARSET_NOT_SUPPORTED, encodingName);
+      throw new GeneratorException();
```
regisd: We could default to platform encoding or utd-8 here

jflex/src/test/java/jflex/AntTaskTests.java
```diff
@@ -163,4 +164,13 @@ public void testLegacyDot() {
     task.setLegacyDot(true);
     assertTrue(Options.legacy_dot);
   }
+
+  public void testSetEncoding() {
+    Charset defaultSet = Charset.defaultCharset();
+    String name = "utf-8";
+    Charset charset = Charset.forName(name);
+    assertTrue(Options.encoding.equals(defaultSet));
```
regisd: Prefer assertEquals

jflex/src/main/java/jflex/Options.java
lsf37: You are right, it'd be better to apply this encoding to both, so that things are consistent. I don't think it's worth doing two separate options, because if somebody wants the output file in a different encoding to the one specified, they can still recode it, as long as it's clear what the output encoding was

jflex-maven-plugin/src/main/java/de/jflex/plugin/maven/JFlexMojo.java
lsf37: yup, will update

jflex/src/main/java/jflex/Options.java
lsf37: I think it's better to fail if the user provided a specific encoding and it's not available. Otherwise it'd be too easy to overlook and get wrong results.

Issue #207, opened 2016-11-05
Print char position in debug mode if char directive is present


mschallnahs:
Equally to the behaviour of `%column` and `%line` when `%debug` is active could `%char` also force to print he value of `yychar `in the debug messages.

lsf37:
That would make sense. Will need to look into if there is any problem with that, sounds like something we should have implemented a long time ago.

lsf37:
This is now implemented in 2b17a9b

lsf37:
Review comments still to address.

PR #309, opened 2018-09-16
add yychar reporting when both %debug and %char are present

lsf37:
The main method of %debug scanners will now report the value of yychar for
each match. As opposed to yyline and yycolumn counting who report yyline+1 and
yycolumn+1, the yychar reporting starts at 0, because it is more likely to be
useful as an index into a buffer (whereas line and column are more likely to
be useful in an editor).

Fixes issue #207.

### DIFF
--- a/jflex/src/main/java/jflex/Emitter.java
+++ b/jflex/src/main/java/jflex/Emitter.java
@@ -278,6 +278,7 @@ public final class Emitter {
       print("    System.out.println( ");
       if (scanner.lineCount) print("\"line:\" + (yyline+1) + ");
       if (scanner.columnCount) print("\" col:\" + (yycolumn+1) + ");
+      if (scanner.charCount) print("\"char:\" + yychar + ");
       println("\" --\"+ yytext() + \"--\" + getTokenName(s.sym) + \"--\");");
       println("    return s;");
       println("  }");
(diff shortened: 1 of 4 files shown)


jflex/src/main/java/jflex/Emitter.java
```diff
@@ -278,6 +278,7 @@ private void emitMain() {
       print("    System.out.println( ");
       if (scanner.lineCount) print("\"line:\" + (yyline+1) + ");
       if (scanner.columnCount) print("\" col:\" + (yycolumn+1) + ");
+      if (scanner.charCount) print("\"char:\" + yychar + ");
```
regisd: white space missing

jflex/src/main/java/jflex/Emitter.java
```diff
@@ -1199,6 +1201,7 @@ private void emitEOFVal() {
             print("              System.out.println(");
             if (scanner.lineCount) print("\"line: \"+(yyline+1)+\" \"+");
             if (scanner.columnCount) print("\"col: \"+(yycolumn+1)+\" \"+");
+            if (scanner.charCount) print("\"char:\" + yychar + ");
             println("\"match: <<EOF>>\");");
```
regisd: white space missing

jflex/src/main/java/jflex/Emitter.java
```diff
@@ -1221,6 +1224,7 @@ private void emitEOFVal() {
         print("                System.out.println(");
         if (scanner.lineCount) print("\"line: \"+(yyline+1)+\" \"+");
         if (scanner.columnCount) print("\"col: \"+(yycolumn+1)+\" \"+");
+        if (scanner.charCount) print("\"char:\" + yychar + ");
```
regisd: white space missing

lsf37:
Good point, will update.

jflex/src/main/java/jflex/Emitter.java
lsf37: In this case, the previous output has a whitespace after, so this one should be fine.

lsf37: but there is one missing afterwards. Ok, will add.

Issue #194, opened 2016-02-27
jflex --{version,info,help} should exit with code 0 (success)


asmod3us:
Title really says it all. In a bash shell, 
```
jflex --version
echo $?
```

will display the version and 1, i.e. a non-zero exit code. Non-zero exit codes usually indicate an error (see e.g. `grep --version; echo $?`), so it would be more idiomatic to exit with code 0.

lsf37:
Yes, it would make a lot of sense to do that. Have put it on the list of things to do for the next release.

PR #318, opened 2018-09-16
Let Travis update the aggregated-java-sources branch

regisd:
We now have a branch [aggregated-java-sources](https://github.com/jflex-de/jflex/tree/aggregated-java-sources).

With this change, Travis will push the java sources of jflex (incl. generated-sources) on every commit on **master**.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,17 +1,24 @@
 *~
+*.versionsBackup
+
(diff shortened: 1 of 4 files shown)


regisd:
N.B. The Travis status is red because of #194

PR #320, opened 2018-09-16
bump version: JFlex 1.7.0-SNAPSHOT –> 1.7.0

regisd:
Change automatically generated by prepare-release.pl


regisd:
There is in fact no reason to do that on master.

lsf37:
Thanks! I'm hoping to be able to get back to the release tonight and will merge then. I want to do the snapshot version update as the last commit, so will rebase this one when the rest is done.

regisd:
This should be the last thing before cutting the release for 1.7.0.
i.e. This should be the last open issue in https://github.com/jflex-de/jflex/milestone/10

lsf37:
I've rebased the branch to make sure I'm not overlooking something with all the merge commits. (Confirmed zero diff to previous version).

testsuite/testcases/src/test/cases/java/java-2.input
```diff
@@ -1,6 +1,6 @@
 /* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * *
- * JFlex 1.4.1                                                             *
- * Copyright (C) 1998-2004  Gerwin Klein <lsf@jflex.de>                    *
+ * JFlex 1.7.0                                                             *
```
lsf37: That change might lead to confusion down the track. The test input is just a file picked from 1.4.1 back then, i.e. updating the version number in it will lead us down the wrong track if we ever want to know where it came from. Are you happy for me to amend/revert this bit?

testsuite/testcases/src/test/cases/java/java-2.input
```diff
@@ -1,6 +1,6 @@
...
+ * Copyright (C) 1998-2018  Gerwin Klein <lsf@jflex.de>                    *
```
lsf37: probably same for this one

PR #321, opened 2018-09-16
Update release instructions

regisd:
* Sonatype requires javadoc and sources.
  Update the `RELEASE.md`.
* Update the post-release.pl script
  Instead of pushing the new change directly on master (which is how protected on github)
  create a new **new_snapshot** branch.


regisd:
N.B. We could do it automatically in the package step, but I don't think it's worth doing it.
https://maven.apache.org/plugin-developers/cookbook/attach-source-javadoc-artifacts.html

lsf37:
Thanks for these!

PR #323, opened 2018-09-17
Minor: Udpate Windows batch script

regisd:
Use `%*` for remaining args, instead of `%1` `%2` `%3` ... `%9`
(which might me a couple)


regisd:
I don't have a Windows machine to test the change, though.

lsf37:
I have this commit on the release branch and I have access to a Windows machine. Should be able to test it tomorrow.

lsf37:
This seems to be working (I tested it in isolation on a windows box). Happy enough to include it in master for now, I'll do a bigger test when I have the full release package working.

regisd:
I'm not using Windows. Up to you if you want to include in 1.7.0 or wait for a subsequent release.

PR #325, opened 2018-09-17
Use Maven wrapper in deploy script

### LINKED ISSUE #315: sonatype deployment broken
It looks like a few things changed slightly since the last time we deployed a release on sonatype.

After a few smaller updates (setting repo URL), and after a successful `./mvn-deploy` run, the sonatype website complains about the following when I stage the release:
  - missing javadoc
  - missing sources
  - missing gpg signature

I think I can address the gpg myself, but I'm a bit lost at the other two artefacts, because I thought we do produce them already. The might just be tweak needed in one of the `pom.xml` files, but without digging into them much more deeply, I couldn't see what exactly.

regisd:
- Prefer `mvnw` over `mvn`
- Also build the javadoc and source jars before deploy. Fix #315

### DIFF
--- a/mvn-deploy.sh
+++ b/mvn-deploy.sh
@@ -43,5 +43,7 @@ cd ../jflex-maven-plugin
 
 # Deploy the deployable stuff to the Sonatype OSS Maven repository
 cd ..
-mvn deploy
+./mvnw javadoc:jar
+./mvnw source:jar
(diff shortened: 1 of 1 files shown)


lsf37:
I'm happy with these changes. Will sync with master and merge if that's fine with you.

PR #328, opened 2018-09-17
Build the doc automatically [WIP]

regisd:
- The Markdown is the only source of truth. Delete LaTeX doc.
- WIP: Generate PDF from md
  - There are unsupported syntax, eg `[^1]` (footnote?)
- TODO: Build PDF on Travis when in a release branch
- TODO: Deploy PDF to web site


lsf37:
Um, these files are not unused, and the PDF generated directly from Markdown looks way too ugly to distribute. The tex files are not duplicating anything, they are either generated from md or are the document frame. I'd really like to leave all of these as is (apart from fixing your name, of course :-))

regisd:
Right. So another solution is to keep the current pipeline with the real LaTeX implementation.
https://tex.stackexchange.com/questions/398830/how-to-build-my-latex-automatically-using-travis-ci
I'm more and more thinking we should have a permanent "stable"/"release" branch which triggers:
- build doc
- build maven site
- deploy all to jflex-web

lsf37:
It'd certainly make sense to build the docs automatically.

Not quite sure about a permanent release branch. Would it really add that much? The (non-)frequency of our releases means that something likely will have broken in between. On the other hand, if it encourages us to make smaller releases more often, maybe that'd be a good thing. I don't have strong feelings either way.

regisd:
Abandoned in favour of #422

PR #333, opened 2018-09-18
Minor: XML codestyle on POM

regisd:
- Make the <project> and its ns declarations on one line
- Add xml declaration on all files


regisd:
@lsf37 This seems unrelated to the actual version bump. Consider merging before #320

PR #342, opened 2018-09-21
Bump to version 1.7.1

lsf37:
Contains two tweak commits to fix up release and post release script for next time


lsf37:
@regisd the `aggregated-java-sources` deployment seems not to have been triggered by the merge of 1.7.0 to master. Not sure if you prefer this merged first or holding off until that is resolved.

regisd:
Yes, I think I decide to trigger the aggregated-java-sources on a branch named "release". The reason is that you probably built the release from jflex-1.7.0 branch ; and I realised that there could be differences with master.
So, in theory, the only thing to do is
```
git co -b release
git push
```
and then the release branch can be deleted too.

lsf37:
Ok, will try that. Otherwise the release is now out and seems to have gone well. I'll let you have a look before announcing it!

regisd:
Ah, my mistake. This change was in PR #332

regisd:
Right. So, I think the Travis config is now correct, and https://travis-ci.org/jflex-de/jflex/builds/431386863 should update the maven-site.
The sources will be updated on evety commit on master, and need to be tagged manually.

lsf37:
There usually are bugs :-) We have a few small things on the 1.7.1 milestone. If they become too big for a minor release we can still bump it up to 1.8.0 fairly easily.

lsf37:
I’ll merge this one now so the repo is back in dev mode. If you feel strongly about going to 1.8.0 directly I’m fine with bumping it up again.

PR #345, opened 2018-09-21
Build maven-site only on branch release

regisd:
- Only deploy the  **maven-site** on branch **release**.
  This is to avoid the documentation to be about -SNPASHOT versions.

TODO: Consider if we should use a tag rather than a branch.


regisd:
Ran ./scripts/deploy-sources.sh manually on this branch.

PR #350, opened 2018-09-21
Read Travis GitHub key to deploy sources

regisd:
See https://docs.travis-ci.com/user/private-dependencies/
Use
 - deploy key
 - or ssh https://github.com/alrra/travis-scripts/blob/master/docs/github-deploy-keys.md


regisd:
Pull requests cannot be tested properly because the secure variables are not set by Travis for obvious security reasons.

PR #353, opened 2018-09-21
Add CUP license in aggregate-java-sources

regisd:
Since we redistribute the CUP runtime, I think we should show the CUP license in the LICENSE flle.


regisd:
Actually, this is not needed when we only ship jflex.
It's only needed for the shaded jar or for the aggregate-java-sources

lsf37:
Yes, copying it ver to the aggregated sources is good. I’ll also add something to the release package script when this is merged, so we’re consistent.

regisd:
OK, submitting. Hopefully Travis will correctly push changes in aggregated-java-sources

PR #362, opened 2018-09-22
 Mavenify cup-interpreter

regisd:


lsf37:
Sorry to be a drag on the quest to mavenify the examples, but I liked the fact that the examples have build system diversity. It makes sense for us to pick one build method internally, but we should show people multiple ways to build, not only one (and esp not only the most complex one). 

This also forces us to make sure we don’t accidentally break people's builds if they don’t run maven.

regisd:
For an example built with ant, the **java** is still ant based.

regisd:
I agree it would be nice to have more build tool examples, for instance with gradle or bazel.

regisd:
See PR #385 instead. It manifies the file structure but keeps multiple build systems.

PR #363, opened 2018-09-22
Add openjdk9 in the build matrix

regisd:
This require compiling for java 1.6 instead of java 1.5 like previously, because openjdk9 doens't support `-source 1.5` anymore.


regisd:
Now, I don't understand. The produce artifact are already for Java 1.7
```
jar xf ~/Downloads/jflex-1.7.0.jar jflex/Timer.class
file jflex/Timer.class
jflex/Timer.class: compiled Java class data, version 51.0 (Java 1.7)
```

PR #364, opened 2018-09-22
Re-enable check against FallThrough violations

### LINKED ISSUE #239: Enable error-prone FallThrough test
In v. 1.7.0 the generated code doesn't have FallThrough violations anymore (see #222).

Re-enable the test as an error (currently downgraded as warning)

regisd:
Fix #239

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -431,9 +431,6 @@
             <configuration>
               <compilerId>javac-with-errorprone</compilerId>
               <forceJavacCompilerUse>true</forceJavacCompilerUse>
-              <compilerArgs>
-                <arg>-Xep:FallThrough:WARN</arg>
-              </compilerArgs>
               <source>7</source>
               <target>7</target>
             </configuration>


regisd:
IIRC, this can only be done with code generated with 1.8.0, hence can only be done in 1.9.0

lsf37:
> IIRC, this can only be done with code generated with 1.8.0, hence can only be done in 1.9.0

Makes sense, let's leave this here until 1.8.0 is out.

lsf37:
This seems to be passing even though bazel is still building with 1.7.0. Is this Ok to merge?

regisd:
I'm afraid not. I tried to update to jflex-1.8 in the corp repository, and I get a number of errors. Sorry that I didn't take the time to investigate.

lsf37:
No worries. Are the errors about fall-through or other jflex 1.8.0 things?

regisd:
So, the scope of the suppress warning (either via `@SuppressWarnings` or via javacop `-X:ep` is only about FallThrough. I'm running a presubmit now.

regisd:
There are remaining violations on 1.8.0. 

```
jflex/java/jflex/core/LexScan.java:2483: error: [FallThrough] Execution may fall through from the previous case; add a `// fall through` comment before this line if it was deliberate
            case 800: break;
            ^
    (see http://go/bugpattern/FallThrough)
```

PR #370, opened 2018-09-24
doc: Bump version of the jflex-maven-plugin to 1.7.1-SNPASHOT

regisd:
The manual was still referencing version 1.6.0


regisd:
Actually not, the Makefile for the doc repalces `$VERSION` automatically.
The pre-release has nothing to do if th edoc uses `$VERSION` instead.

PR #377, opened 2018-09-25
Remove openjdk7 from the build matrix

regisd:
- This will allow to bump the version of multiple maven plugins
- The code is still compiled with `-target 1.7` defined in maven-compiler-plugin
- JDK7 is largely deprecated in the industry


regisd:
Yes, that's correct. The build time dep would be java 8 (which allows to use more recent version of all plugins), but the actual jflex artifacts are still java 7.
The only difference is for java_cup which was currently built for java 5 and that I'm bumping to java 6 because openjdk 9 doesn't support building for java 5 anymore.

PR #388, opened 2018-09-29
Add licenses in jflex-full.jar

regisd:
Add `LICENSE_JFLEX` in JFlex jar and `LICENSE_CUP` in java_cup_runtime.

The shaded jflex-full-$VERISON.jar hence contains both licenses.

Roll forward fix of #386 
- symlinks in git are just the path of the file
- instead, place the actual LICENSE file in `src/main/resources` and add a symlink in the root dir for convenience


regisd:
Git lost track of moved files

PR #396, opened 2018-09-30
Travis: drastically reduce the build matrix.

regisd:
We have useless matrix expansion (e.g. java-format un JDK8 and java-format in JDK9).

Instead of expanding all possible values, explicitely define the inclusion we want.


regisd:
before  https://travis-ci.org/jflex-de/jflex/builds/435236868
after https://travis-ci.org/jflex-de/jflex/builds/435262519

PR #404, opened 2018-10-03
Travis: Use language minimal for bazel

regisd:
> The minimal image contains:
> - version control tools
> - essential build tools such as gcc and make
> - network tools such as curl and essential Docker
https://docs.travis-ci.com/user/languages/minimal-and-generic/


regisd:
Done in $406

PR #414, opened 2018-10-06
Don't infer package name from lex rule names.

### LINKED ISSUE #104: [Bug] Generating java class in strage folder [sf#102]
**Reported by *anonymous on 2009-03-26 15:35 UTC**
Hello everyone\!

I'm having a rather strange problem with JFlex's plugin for maven. It is generating my java source file containing the regular expression in hierarchy of folders which are named "Name = yytext\(\)" \ "substring\(0,yytext\(\)" \ "length\(\)\)".

Has anyone been through this before? I'm gonna attatch the POM file I'm using for the project.

regisd:
Fix #104

Also, make ClassInfo immutable.

### DIFF
--- a/jflex-maven-plugin/pom.xml
+++ b/jflex-maven-plugin/pom.xml
@@ -80,6 +80,7 @@
     <dependency>
       <groupId>com.google.code.findbugs</groupId>
       <artifactId>jsr305</artifactId>
+      <version>3.0.2</version>
     </dependency>
     <dependency>
       <groupId>com.google.guava</groupId>
--- a/jflex-maven-plugin/pom.xml
+++ b/jflex-maven-plugin/pom.xml
@@ -78,6 +78,10 @@
   </build>
   <dependencies>
     <dependency>
+      <groupId>com.google.code.findbugs</groupId>
+      <artifactId>jsr305</artifactId>
+    </dependency>
+    <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
     </dependency>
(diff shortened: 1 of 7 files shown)


lsf37:
This will still fail if the string `package` is mentioned on the same line as a semicolon in later sections of the lexer spec.

Test case:

```java
  @Test
  public void guessPackageAndClass_givenClass_package_in_rule() throws Exception {
    String lex =
        "\n"
            + "%%\n"
            + "\n"
            + "%public\n"
            + "%class Foo\n"
            + "\n"
            + "%apiprivate\n"
            + "%int\n"
            + "\n"
            + "%%\n"
            + "\n"
            + "package X  { /* no action; */ }"
            + "\n"
            + "[^]  { /* no action */ }\n";
    assertThat(guessPackageAndClass(lex)).isEqualTo(new ClassInfo("Foo", null));
  }
```

The code should probably become more aware in which section it currently is, and scan for package name only up until the first `%%` and for class name only after the first `%%` until the second `%%`.

regisd:
> This will still fail if the string `package` is mentioned on the same line as a semicolon in later sections of the lexer spec.
> 
> Test case:
> 
>             + "\n"
>             + "package X  { /* no action; */ }"
>
> The code should probably become more aware in which section it currently is, and scan for package name only up until the first `%%` and for class name only after the first `%%` until the second `%%`.

You are absolutely right. There are 2 reasons why I dont' want to go for a proper fix:

- This only happens if the package is not defined in the Java user code ; which is never the case for production code ; i.e. we are only fixing examples.
- Changes may very well be confllcting with my refactoring of Out and Options.

In the end, there are only three robusts solutions:
- Only give the output directory ; but this breaks a lot of abstraction. I think the Emitter shuold only have an OutputStream (which is convenient for tests)
- Let the lexer do two passes.
- Output the java code in a temp file, and move it at the end of the generation, once the emitter can tell exactly what package it is

regisd:
I'll submit this for now.

lsf37:
Yes, that's alright. I'll make a low-priority issue so I don't forget coming back to this once you've merged the Out/Option refactor. (I think I can fix this properly reasonably easily).

Issue #390, opened 2018-09-29
 Failure to find de.jflex:cup_runtime:jar:11b


ambs:
Hi. I'm new to Java and all these new ways to automate builds. I am trying to build a simple project in IntelliJ IDEA. I configured Maven as described in the documentation. Nevertheless, when running the compile step, I get 

org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal de.jflex:jflex-maven-plugin:1.7.0:generate (default) on project pt-flex: Execution default of goal de.jflex:jflex-maven-plugin:1.7.0:generate failed: Plugin de.jflex:jflex-maven-plugin:1.7.0 or one of its dependencies could not be resolved: Failure to find de.jflex:cup_runtime:jar:11b in https://repo.maven.apache.org/maven2 was cached in the local repository, resolution will not be reattempted until the update interval of central has elapsed or updates are forced

Any help is welcome.

regisd:
Our mistake, we didn't publish **de.jflex:cup_runtime**.
@lsf37 I don't have permissions to do this, so I'm afraid this problem falls on you.

regisd:
@ambs Thanks a lot for reporting the problem! We didn't deploy one of our internal dependency.

I see two options to use JFlex on your project
- use jflex version 1.6.1 (instead 1.7.0) until this bug is fixed.
- install jflex from sources:
  ```
    git clone --branch v1.7.0 --depth 1 https://github.com/jflex-de/jflex.git 
    cd jflex
    ./mvnw install
  ```

ambs:
Thanks. Will try 1.6.1 for now, as it is easier to explain to students.

regisd:
Let's follow-up in #400

lsf37:
Sorry about all this, I'm looking into it now. It seems the pom file for cup explicitly tells it to skip deployment, but I guess that should only apply to cup itself and not the runtime.

lsf37:
So, I've manually deployed cup_runtime and cup-parent 11b. For some reason maven still does not find cup-parent even though it does find and download the pom and the jar for cup_runtime.

Maybe it just needs some more time to sync, but why it would sync only one and not the other mystifies me. Will give it a bit more time.

lsf37:
And now it's there and seems to work.

The reason I didn't catch this in my release testing is that maven caches things locally, which means all dependencies were present on my machine and it didn't try to download them.

regisd:
[...]

PR #415, opened 2018-10-07
deploy cup_runtime and cup-parent

lsf37:
This should make sure #390 doesn't happen. Unfortunately this duplicates some info from jflex-parent, but we probably don't want cup-parent to depend on jflex-parent.

### DIFF
--- a/cup/cup/pom.xml
+++ b/cup/cup/pom.xml
@@ -17,6 +17,13 @@
     <plugins>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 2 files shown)


regisd:
Thanks! Actually, I think cup should depend on the jflex-parent more than it does today. For instance `mvn site` creates a broken site because it links to pages on http://www2.cs.tum.edu/projects/cup/ which don't exist.

regisd:
Also, I know you're not a fan of squash commits, but I think they are much neater than merge commits. In fact, when there is only one commit in the PR, I like the rebase commit too.

PR #423, opened 2018-10-07
Add owners file

regisd:
This will:

- protect the cup directory
  As a project admin, you can still commit. But the merge button is red.
  https://help.github.com/articles/about-codeowners/
- help github in suggesting reviewers

PR #424, opened 2018-10-08
Test change in cup_runtime.

regisd:
Do not submit :-)


regisd:
Ignore.
This was a test for #423

Issue #214, opened 2017-04-13
make the generator thread safe


lsf37:
As discussed in issue #211 and PR #213, the JFlex generator is currently not thread safe. It'd be nice for larger projects if multiple concurrent build invocations could be supported.

Main thing to change would be the static setup of the `Options` class, I think.

regisd:
Interesting project. I'll give it a try.

PR #428, opened 2018-10-10
Split LexGenerator out of Main

regisd:
This is a minor refactoring to support multithreading  #214

Split Main in 2 classes:

* `Main` the cli client that parses the command line options
* `LexGenerator` that generates java code
   The later is used by the **cli**, the **jflex-maven-plugin**, the **ant task** and the **GUI**, i.e. `Main.generate(lexFile)` becomes `LexGenerator().generate(lexFile)`

In the future, `LexGenerator.generate()` will not be static and have internal state in order to support multithreading.


jflex/src/main/java/jflex/LexGenerator.java
```diff
@@ -0,0 +1,131 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+
+/**
+ * This is the main class of JFlex controlling the scanner generation process. It is responsible for
+ * parsing the commandline, getting input files, starting up the GUI if necessary, etc.
```
lsf37: The comment will need adjustment for the refactor, indicating which part is in which class in the new setting.

PR #432, opened 2018-10-10
Remove ant build

regisd:
Remove the ant build for JFlex. In my opinion, it adds maintenance cost but provides no benefit.
What's more, it can mislead someone into thinking that the .jar we publish is made with ant.

Note that ant is still used on the examples ; which is valuable.


regisd:
I don't think there  is any reference in the documentation (except the JFlexAntTask or the examples).
But there was still hack in `mk-release.sh`. I removed it.

lsf37:
Good catch. LGTM, then.

PR #439, opened 2018-10-12
Build Manual with Bazel

regisd:
* Build the user manual with Bazel
  - This is WIP. I disabled the styles to have something working
  - Build the aggregated markdown with variable substitution with a simple genrule
  - Use rule bazel-pandoc rule to build the html and the latex formats https://github.com/ProdriveTechnologies/bazel-pandoc
  - Actually my fork until they allow to specify the output file https://github.com/ProdriveTechnologies/bazel-pandoc/pull/1
  - Use latex-rule to build the PDF https://github.com/ProdriveTechnologies/bazel-latex
  - Do stupid latex concatenation until the latex rule supports \include{} or pandoc template https://github.com/ProdriveTechnologies/bazel-latex/issues/3
* Note that the VM for Bazel doesn't install (with apt) pdflatex or pandoc: Everything is self contained by the rules. We need to update the rules to update the toolchain, but that means the build is reliable and consistent.
* Update documentation


regisd:
So you can try

```
bazel build //docs:html
bazel build //docs:manual
```

Issue #429, opened 2018-10-10
examples/simple ant run fails with ExecuteJava can not access a member of class Yylex with modifiers "public static"


regisd:
Steps:
```
cd jflex/examples/simple
ant run
```

Expected:
Execute Yylex

Actual:
> build:
>     [javac] /Users/regis/Projects/jflex/jflex/examples/simple/build.xml:25: warning: 'includeantruntime' was not set, defaulting to build.sysclasspath=last; set to false for repeatable builds
> 
> run:
>      [java] java.lang.IllegalAccessException: Class org.apache.tools.ant.taskdefs.ExecuteJava can not access a member of class Yylex with modifiers "public static"
>      [java] 	at org.apache.tools.ant.taskdefs.ExecuteJava.execute(ExecuteJava.java:194)
>      [java] 	at org.apache.tools.ant.taskdefs.Java.run(Java.java:861)
>      [java] 	at org.apache.tools.ant.taskdefs.Java.executeJava(Java.java:231)
>      [java] 	at org.apache.tools.ant.taskdefs.Java.executeJava(Java.java:135)
>      [java] 	at org.apache.tools.ant.taskdefs.Java.execute(Java.java:108)
>      [java] 	at org.apache.tools.ant.UnknownElement.execute(UnknownElement.java:292)
>      [java] 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
>      [java] 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
>      [java] 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
>      [java] 	at java.lang.reflect.Method.invoke(Method.java:498)
>      [java] 	at org.apache.tools.ant.dispatch.DispatchUtils.execute(DispatchUtils.java:99)
>      [java] 	at org.apache.tools.ant.Task.perform(Task.java:350)
>      [java] 	at org.apache.tools.ant.Target.execute(Target.java:449)
>      [java] 	at org.apache.tools.ant.Target.performTasks(Target.java:470)
>      [java] 	at org.apache.tools.ant.Project.executeSortedTargets(Project.java:1388)
>      [java] 	at org.apache.tools.ant.Project.executeTarget(Project.java:1361)
>      [java] 	at org.apache.tools.ant.helper.DefaultExecutor.executeTargets(DefaultExecutor.java:41)
[...]

PR #304, opened 2018-09-04
Bazel [WIP]

regisd:
Crazy attempt to replace Maven by Bazel

PR #440, opened 2018-10-12
Move Bazel WORKSPACE to the root directory.

regisd:
- The Bazel workspace should generally be the root. This is what we want for building the docs (#429) or the whole project (#304)
- This allows to reference the jflex.jar and the cup.jar directly; the Bazel script doesn't have to copy them anymore.


regisd:
In Bazel, WORKSPACE only defines what `//` in the rules mean.
It wouldn't be possible to access cup if jflex if it's not in the workspace.
What we ship has really nothing to do with where the Workspace file is.

PR #445, opened 2018-10-13
Start migration of the testsuite to Bazel

regisd:
**Why**
The testsuite currently runs with a Maven plugin and is the slowest part of the build. Migrating to Bazel would be the biggest win for CI, in the case when files outside of jflex (core) are modified.

**How**
- Add a custom junit test runner
- Replace test specifications `*.test` by normal jUnit classes

This is change is a prototype for `apiprivate`


regisd:
This doesn't touch existing code or impact the distribution targz, so I don't think it's controversial. I hope you like the new approach (fully java) to define tests.

PR #447, opened 2018-10-14
New module: Common testing

regisd:
* Maven working directory is where the `pom.xml` is
* Bazel working directory is the sandboxed runfiles directory (which corresponds to the root WORKSPACE).

As a result opening files in test differs.

In the **examples/simple** `YylexTest`, I made a simple `openFile()` method which tunes the path accordingly to the build tool used.

It's time to place this code in a common module, so that all tests can benefit.


regisd:
@lsf37 I finally understand your concerns, and propose to move this discussion in #449 

This change is overly complex anyway. Abandon.

PR #452, opened 2018-10-14
Move testsuite/manual-ex to examples

regisd:
Move the example of the manual:
- from testsuite/cases/manual-ex
- to examples/cup-java-minijava


regisd:
This is to help answer https://stackoverflow.com/questions/52573382/jflex-cup-error-cannot-find-symbol-return-symbolsym-eqeq

regisd:
I also understand why someone would want to use the example of the manual to try JFlex.

Writing documentation is hard, and I've added a few more pointers in external grammars.

PR #462, opened 2018-10-16
New Layout for Aggregated java sources

regisd:
The code is now in directory `/java`.


regisd:
Oh no, I want to merge in agregated-java-sources.

PR #471, opened 2018-10-17
Fix aggregated-java-sources execution

regisd:


regisd:
Arg! Wrong branch again

PR #475, opened 2018-10-18
Move class Timer in jflex.performance

regisd:
- Move java class in new package
- Add bazel package


regisd:
The package could be "timing" or "logging". Anyway, this class is a leaf in the dependency tree, and should be moved in its own package (which depends on nothing as a result)

PR #478, opened 2018-10-19
Create module jflex.gui - split GUI out of core jflex

regisd:
I'm annoyed by the cyclic dependency between the root `jflex` package and `jflex.gui`.

I propose to
- create a build target for the GUI
- don't make 
- defer the decision to run cli / gui to the script

Current deps between pacakges

### DIFF
--- a/jflex/bin/jflex
+++ b/jflex/bin/jflex
@@ -44,7 +44,7 @@ PRG=${PRG%/*}
 JFLEX_HOME=${PRG}/.. 
 
 # --------------------------------------------------------------------
-if [ "$#" -eq 0 ];
+if [ "$#" -eq 0 ]; then
   $JAVA -Xmx128m -jar "$JFLEX_HOME"/lib/jflex-full-${JFLEX_VERSION}.jar jflex.gui.MainFrame
 else;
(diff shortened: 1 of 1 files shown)


regisd:
jflex-packages

regisd:
**After**  Found 146 nodes, with 2 slice edges in violation of dependency constraints.

jflex-packages

lsf37:
I don't think shifting decisions into the scripts is the right direction. For one, it creates duplication of logic, and second it distributes logic over multiple languages, i.e. we'll tend to forget about it when we focus on the Java code. I don't see why that circular dependency is a problem. Packages relationships don't have to be trees.

regisd:
Tree deps are cleaner; it's perfectly fine to have circular dependencies within a package. Much less so between packages.

My root motivation is to be able to ship jflex_core (or lets's say jflex-maven-plugin) _without_ the gui and the ant task.

regisd:
Related to this #486 will improve the separation in terms of bazel build.

lsf37:
If we really want to remove the dependency, then I think we should make a `main` or `cli` package that depends on the generator and the GUI. The GUI then only needs to depend on the generator, and `main` invokes either the generator or the GUI, depending on arguments. Would that make sense?

regisd:
Replaced by #496

PR #491, opened 2018-10-20
[WIP] Bzltestsuite

regisd:


regisd:
Abandonned in favour of #492, #493 and #494

PR #413, opened 2018-10-06
Fix PMD violations

regisd:
From https://lgtm.com/projects/g/jflex-de/jflex/alerts/

- Class Interval overrides equals but not hashCode.
- Class IntCharSet overrides equals but not hashCode.
- StateSet: equals() method does not seem to check argument type.
- Main: array access might be out of bounds, as the index might be equal to the array length.
- Main: This FileInputStream is not always closed on method exit.
- Skeleton: This FileReader is not always closed on method exit.
- Skeleton: The FileInputStream and FileOutputStream classes contains a finalizer method which will cause garbage collection pauses. See JDK-8080225 for details.

PR #495, opened 2018-10-21
Use @AutoValue

regisd:
Roll forward of #433 

`@AutoValue` is a **build-time** dependency that helps to remove boiler-plate code in value objects.

- It improves code health
- This would fix avoid coding mistakes such as #413
- It also makes the value object immutable ; which is another good practice to avoid bugs.

See https://github.com/google/auto/blob/master/value/userguide/index.md

**N.B** AutoValue Release 1.5 requires JDK8 ; hence using old version https://github.com/google/auto/releases/tag/auto-value-1.5


regisd:
Travis is stuck. Replaced by #505

PR #496, opened 2018-10-21
Move all but Main in package jflex.core

regisd:
The dependencies within JFlex are a extremely cyclic

jflex-packages-before

With this change (and previous PR #476 #481 #482), the packages are much better organised

jflex-packages-after


regisd:
There are still possible improvements within `jflex.core` but the previous changes also make them way more visible.
jflex-packages-core

regisd:
Detailed view of the  deps with this change
jflex-packages

PR #501, opened 2018-10-21
Create module jflex.generator

regisd:
@lsf37 I'll let you review this change, because it's quite a major refactoring.

This change is in project [Modularization](https://github.com/jflex-de/jflex/projects/14) and aims at separating
* The JFlex generator (LexGenerator, Emitter) in new package `jflex.generator`
* from JFlex core (the lexer and scanner).

jflex-packages

There are some important changes:
 * The JFlex `VERSION` is used in too many places. I moved it in (new package) `jflex.base.Build`
* The Emitter doesn't access fields of the scanner directly anymore, it uses getters
  * and it particular the emitter doesn't mutate the `LexScan#functionName` anymore ; it has its own field (initialised from the LexScan value)
* The Emitter doesn't access fields of the DFA or NFA directly anymore, it uses getters
* The Emitter still accesses fields in `Action` or `CharClassInterval`, but those fields are now immutable (fields are final)


regisd:
Note that the individual commits are quite clean if you want to see the progress that lead to this change.

regisd:
After the change:
jflex-packages

PR #509, opened 2018-10-22
Revert "Add compile-aggregate-sources.sh"

regisd:
Reverts jflex-de/jflex#508


regisd:
Github "Revert" button is wrong.

PR #511, opened 2018-10-22
[WIP] Build the jflex-maven-plugin with Bazel 

regisd:


regisd:
This is not realistic. Problem this only module that needs to be build by Maven is the maven plugin.

PR #512, opened 2018-10-23
Fix compile.sh on aggregated-java-sources

regisd:
./compile.sh: 14: return: Illegal number: ant-1.7.0.jar


regisd:
Follow-up of #505 and #510

PR #513, opened 2018-10-23
Add @Generated on generated lexer

regisd:


regisd:
Arg! With Java 9 modularization, this breaks the generated code:
>  package javax.annotation is not visible

In Dagger, they solved it with a method that returns `javax.annotation.processing.Generated` for JDK 9 and newer, `javax.annotation.Generated`. See https://github.com/google/auto/commit/0383c1cf7472d7444463353aa2ac956c9471130a

regisd:
This was fixed after https://github.com/jflex-de/jflex/tree/branch-1.7.1 was cut, hence will be in 1.8.0.

PR #526, opened 2018-10-28
Generate UnicodeProperties with Bazel

regisd:
The jflex-unicode-maven-plugin is slow and fetches data from unicode.org which is also not very reliable. As a result, the generated code has been checked in. This is bad practice, and we have been modifying these generated files.

This is the first step in an attempt to replace the jflex-unicode-maven-plugin by Bazel:

1. Bazel fetches and caches all resources. It can use mirrors. See #522
2. With, this change, only `UnicodeProperties` is re-generated, using versions given in the command-line.
3. Instead of using a custom "skeleton", I'm using Apache Velocity template engine. See #520
4. I've rewritten more than I wanted because the URL was too much part of the previous model. See [DataFileType.java](https://github.com/jflex-de/jflex/blob/65c0194e075b0b17c094277a1eb6a59cec49a916/jflex-unicode-maven-plugin/src/main/java/jflex/maven/plugin/unicode/DataFileType.java)

Effective changes in generated file:

* Set default version to 9.0.0 instead of 9.0. This should be a no-op.
* Use switch/case rather than chain of ifs
* Bump unicode 3.1.0 to 3.1.1


regisd:
I'm  wondering why the current UnicodeProperties are generated for all `major.minor` versions. Wouldn't the binding on all the latest major version be sufficient. Why would someone want to use Unicode 2.0.24 rather than 2.1.9?

regisd:
Also, I've repaled the chain of if by a switch/case.
In the end, the overall change can be seen with the golden file of the test: https://github.com/jflex-de/jflex/pull/526/files#diff-d82c6130b238cf8191f0b458e02661da

Alternatively, patch the change and run:

    bazel build  //java/jflex/ucd_generator:gen_unicode_properties
    cp bazel-genfiles/java/jflex/ucd_generator/UnicodeProperties.java jflex/src/main/java/jflex/core/unicode/UnicodeProperties.java

lsf37:
> I'm wondering why the current UnicodeProperties are generated for all `major.minor` versions. Wouldn't the binding on all the latest major version be sufficient. Why would someone want to use Unicode 2.0.24 rather than 2.1.9?

For repeating an older build for instance (in debugging). Might even be ourselves. If they are generated, why remove them?

sarowe:
> On Oct 28, 2018, at 6:06 PM, Gerwin Klein <notifications@github.com> wrote:
> 
> 
>> I'm wondering why the current UnicodeProperties are generated for all major.minor versions. Wouldn't the binding on all the latest major version be sufficient. Why would someone want to use Unicode 2.0.24 rather than 2.1.9?
> 
> For repeating an older build for instance (in debugging). Might even be ourselves. If they are generated, why remove them?

Unicode used to release minor versions with non-trivial changes.  Why not keep them?

In my view, this question is similar to "why bother keeping more than the latest Unicode version compat"?  Part of my motivation in contributing was to allow for people to continue to use older Unicode versions with modern software, separating the two concerns.

Steve

regisd:
I have another question though: is it expected that version 2 is bound with 2.0.14 rather than 2.1.9?

sarowe:
Yes.

The Unicode versioning scheme in JFlex is like this:

2 == 2.0 == 2.0.latest-patch-version
2.1 == 2.1.latest-patch-version

Steve

regisd:
Ok, thanks for the quick answer!
It's actually easier to generate :-)

regisd:
PTAL

sarowe:
I'll have time to review this weekend.

sarowe:
I uninstalled my MacOS Homebrew bazel v0.20 and installed v0.21 via `brew install bazelbuild/tap/bazel`, which appeared to fix the 'CompactStrings' problem.  Now I get a different error:

```
$ bazel run //java/jflex/ucd_generator:Main
INFO: Invocation ID: 28cdfc36-b232-47df-beac-5195af4f3660
ERROR: error loading package '': Encountered error while reading extension file 'repositories.bzl': no such package '@bazel_pandoc//': The native http_archive rule is deprecated. load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive") for a drop-in replacement.
Use --incompatible_remove_native_http_archive=false to temporarily continue using the native rule.
ERROR: error loading package '': Encountered error while reading extension file 'repositories.bzl': no such package '@bazel_pandoc//': The native http_archive rule is deprecated. load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive") for a drop-in replacement.
Use --incompatible_remove_native_http_archive=false to temporarily continue using the native rule.
INFO: Elapsed time: 0.068s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded)
FAILED: Build did NOT complete successfully (0 packages loaded)
```

I tried adding the `--incompatible_remove_native_http_archive=false` option as suggested above, but got the same error: 

```$ bazel run //java/jflex/ucd_generator:Main --incompatible_remove_native_http_archive=false
INFO: Invocation ID: 349a0d1b-4d62-4d6b-9c86-efde9f67fc96
INFO: SHA256 (https://github.com/ProdriveTechnologies/bazel-latex/archive/v0.9.tar.gz) = ecab535bb50699817cea662014432b4398edd7318ee6a6f64399a3287010961c
ERROR: /private/var/tmp/_bazel_sarowe/907433835c955c9dc631764523d6be53/external/bazel_pandoc/BUILD.bazel:7:1: in toolchain_type attribute of toolchain rule @bazel_pandoc//:pandoc_toolchain_linux: rule '@bazel_pandoc//:pandoc_toolchain_type' does not exist. Since this rule was created by the macro 'pandoc_toolchain', the error might have been caused by the macro implementation in /private/var/tmp/_bazel_sarowe/907433835c955c9dc631764523d6be53/external/bazel_pandoc/toolchain.bzl:30:26
ERROR: While resolving toolchains for target //java/jflex/ucd_generator:Main: invalid registered toolchain '@bazel_pandoc//:pandoc_toolchain_linux': Analysis of target '@bazel_pandoc//:pandoc_toolchain_linux' failed; build aborted
ERROR: Analysis of target '//java/jflex/ucd_generator:Main' failed; build aborted: Analysis of target '@bazel_pandoc//:pandoc_toolchain_linux' failed; build aborted
INFO: Elapsed time: 4.087s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (8 packages loaded, 24 targets configured)
FAILED: Build did NOT complete successfully (8 packages loaded, 24 targets configured)
```

Any ideas how to fix this @regisd ?

regisd:
Apologies for the slow response too.

> Many of the new files you've added here include the Apache license, but JFlex is BSD licensed

True. Google preferred license is Apache. If this is incompatible with BSD, I think it might be OK to use BSD since that is used by the rest of the project.

> The JFlex Maven Unicode plugin automatically finds new versions and new point versions, where the Bazel setup instead requires manual intervention to name the exact version(s) to be included and described (e.g. the SHA).  This is probably fine, in that it's very unlikely that there will be new point releases of old Unicode versions, or even point releases of modern releases, given the modern Unicode release cadence.

That's correct, and impossible to work around. In Bazel philosophy, the build must be reliable, hence the exact inputs must be specified.

I just added Unicode 10 to align on master, and the boilerplate is very limited.

> I could not figure out how to invoke the Bazel functionality. (I have zero Bazel experience.) What commands should I type from which directory? (It would be very helpful to add this kind of info somewhere in the source, like in the README.md files.)

Apologies for the lack of doc. Added to README, and this is done with (but you shouldn't have to run this manually, it's just a build dep of the main jflex target ; it's only useful to analyze the generated file as we discuss this refactoring)

```
bazel build java/jflex/ucd_generator:gen_unicode_properties
```

regisd:
> ERROR: While resolving toolchains for target //java/jflex/ucd_generator:Main: invalid registered toolchain '@bazel_pandoc//:pandoc_toolchain_linux'
I've updated this toolchain in #548 
Let me know if it now works for you too.

regisd:
I'll submit the generator now, because it is hermetic to the rest of the codebase.
I'll revert UnicodeProperties itself.

regisd:
I'll submit the generator now, because it is hermetic to the rest of the codebase.

I'll revert UnicodeProperties itself to have zero impact. You can see the changes I introduced in the template in commit 54c67ff

PR #547, opened 2019-02-12
Use bazel image rather than Cirruslab image

regisd:
This should work around cirruslabs/docker-images-bazel/issues/3


regisd:
Submitted in larger PR #548

PR #557, opened 2019-03-10
Refactor unicode Maven plugin

regisd:
Extract a common interface that will be used by both the Maven-based jflex-unicode-maven-plugin and the bazel-based ucd_generator.

The goal is to reuse the Unicode flex scanners as much as possible.


regisd:
I abandon the idea. With Bazel, I want to get rid of the jflex-unicode-maven-plugin instead.

PR #558, opened 2019-03-12
make yychar long

### LINKED ISSUE #536: Lexer has negative yychar when processing files larger than 2 GiB
The `yychar` variable that holds file offset is `int`, therefore limiting the length of files to 2 GiB - 1. We bumped into this in https://github.com/oracle/opengrok/issues/2560

sarowe:
make yychar long to enable scanners with the %char option to consume files larger than 2GB.  Closes #536


sarowe:
Hi @lsf37 or @regisd I'd appreciate a review before I merge.

jflex/changelog.md
```diff
@@ -1,6 +1,7 @@
 ## [JFlex 1.8.0](https://github.com/jflex-de/jflex/milestone/11)
 - Unicode 10.0 and 11.0 are supported (#540, #555)
 - Unicode Emoji properties are supported for Unicode versions 8.0+ (#546)
+- make yychar long to enable scanners with the %char option to consume files larger than 2GB (#536)
```
regisd: nit: you could also format yychar and %char was coe with \`\`

testsuite/testcases/src/test/cases/over-2gb-input/.gitignore
```diff
@@ -0,0 +1 @@
+Over2GbInput.java
```
regisd: Generated files shouldn't be in `src/test`

regisd: Also, I think you didn't intent to submit `Over2GbInput-0.input`

sarowe: This file is the over-2GB input file generator; it is not itself generated.

sarowe: I did mean to, actually.  This is the input file that is over 2GB.  It would be better to generate it at test time, though that slow the test down quite a bit.  I'll look into generating at test time.

regisd: OK, I think such a large file in git is not a good idea. I think a quick workaround is to zip that file, and open the zipped version in the test.

regisd:
Hmmm, I've been trying to reproduce the bug, and if the consumer doesn't read yychar, then nothing bad happens.
Are you sure the given test would fail today?

regisd:
Please see #605

PR #572, opened 2019-11-05
add factory class for Emitter

VA-Wizard:
add new constructor to Emitter class

### DIFF
--- a/jflex/src/main/java/jflex/generator/Emitters.java
+++ b/jflex/src/main/java/jflex/generator/Emitters.java
@@ -1,0 +1,66 @@
+/* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * *
+ * JFlex 1.8.0-SNAPSHOT                                                    *
+ * Copyright (C) 1998-2018  Gerwin Klein <lsf@jflex.de>                    *
+ * All rights reserved.                                                    *
+ *                                                                         *
+ * License: BSD                                                            *
+ *                                                                         *
+ * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
+
+package jflex.generator;
+
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStreamWriter;
+import java.io.PrintWriter;
+import jflex.core.DFA;
+import jflex.core.LexParse;
+import jflex.core.Options;
+import jflex.core.Out;
+
+/** Factory class for Emitter */
+public final class Emitters {
+
+  private Emitters() {}
+
+  /**
+   * Create Emitter that writes to file
+   *
+   * @param inputFile input grammar.
+   * @param parser a {@link LexParse}.
+   * @param dfa a {@link DFA}.
+   * @return {@link Emitter}.
+   * @throws java.io.IOException if any.
+   */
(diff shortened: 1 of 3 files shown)


VA-Wizard:
> Hi @VA-Wizard, what's the purpose of this new constructor? It's not used anywhere else in the code.

Hi @lsf37 , 
Emitter class is high coupled with other classes in the package and provides the actual java code, so that I cannot use its logic outside the package
It's extremely hard to use it without creating any files

I added it there as if it is a default constructor to get more control of the output

the main purpose was to create reloadable java classes in runtime without manipulating the files
you can see the example here: https://github.com/VA-Wizard/jFlex-reloadable

regisd:
LGTM. This constructor looks quite natural to me. Ideally, we would take a `Skeleton` too, but the usage of `Out` makes everything more complicated there.

regisd:
Actually, I propose to go one step further:

* have a single constructor (dumb, which only sets its fields) looking very similar to the one added here.
* have factory methods `Emitters.createFileEmitter` and `Emitters.createPrintWriterEmitter()`.

VA-Wizard:
The factory class seems to be better

jflex/src/main/java/jflex/generator/Emitters.java
```diff
@@ -0,0 +1,66 @@
...
+  /**
+   * Create Emitter that writes to file
+   *
+   * @param inputFile input grammar.
+   * @param parser a {@link LexParse}.
+   * @param dfa a {@link DFA}.
+   * @return {@link Emitter}.
+   * @throws java.io.IOException if any.
+   */
+  public static Emitter createFileEmitter(File inputFile, LexParse parser, DFA dfa)
```
regisd: Please rename `inputFile` to `grammarFile` or `lexFile` or `inputLexFile`.

jflex/src/main/java/jflex/generator/Emitters.java
```diff
@@ -0,0 +1,66 @@
...
+import java.io.PrintWriter;
+import jflex.core.DFA;
+import jflex.core.LexParse;
+import jflex.core.Options;
+import jflex.core.Out;
+
+/** Factory class for Emitter */
+public final class Emitters {
+
+  private Emitters() {}
```
regisd: Good practice. Thanks!
(no action needed)

jflex/src/main/java/jflex/generator/Emitters.java
```diff
@@ -0,0 +1,66 @@
...
+
+  /**
+   * Create Emitter that writes to file
```
regisd: Please make it explicit what file receives the java code. Something like:

> Create an Emitter that generates the java code in a file. The output file name is inferred from the class defined in the grammar.

jflex/src/main/java/jflex/generator/Emitters.java
```diff
@@ -0,0 +1,66 @@
...
+   *
+   * @param inputFile input grammar.
+   * @param parser a {@link LexParse}.
+   * @param dfa a {@link DFA}.
+   * @return {@link Emitter}.
+   * @throws java.io.IOException if any.
```
regisd: nit: no need for the fqn since the class is imported

jflex/src/main/java/jflex/generator/Emitters.java
regisd: nit: in the google code style, this would be "Creates ..."
(no action needed, jflex doesn't really have a code style)

PR #581, opened 2019-11-16
fix negation bug

### LINKED ISSUE #567: r|!r does not always match everything
Original bug report:

hello,
  I found an issue with the negation operator "!"
  With the following specification, string "baba" is not matched
  by either EXP ou !EXP .

Pascal Hennequin

```jflex
%%
%standalone
%{
void ECHO(String cat) { System.out.print("["+cat+":"+yytext()+"]"); }
%}

EXP = ( [^a] [^]* [^a] )
ALL = {EXP} | ! {EXP}  

%%
{ALL}  { ECHO("1"); }
baba   { ECHO("2"); }
```

lsf37:
fixes #567


lsf37:
@regisd @sarowe I'm removing the Java 7 API check for jflex in 822036c90f6628b7207099b178dbb7fa207fa14a.

Given that Java 7 is no longer supported, I think we can now drop it. Any concerns?

sarowe:
> @regisd @sarowe I'm removing the Java 7 API check for jflex in [822036c](https://github.com/jflex-de/jflex/commit/822036c90f6628b7207099b178dbb7fa207fa14a).
> 
> Given that Java 7 is no longer supported, I think we can now drop it. Any concerns?

Works for me.

PR #582, opened 2019-11-16
UCD generator

regisd:
Step 1: Generate empty Unicode_x_y.java (done in #554)
Step 2: Emit maxUnicodeValue in Unicode_x_y.java (with Bazel build).


lsf37:
The `.bazelrc` conflict is probably my fault, I set `build --javacopt "-Xep:Java7ApiChecker:OFF"` instead of removing the line. Will leave it to you do decide which is better..

Issue #216, opened 2017-06-15
Macros in character classes are not supported


dscorbett:
Using a macro in a character class produces the error message “Macros in character classes are not supported.” This contradicts [the documentation](http://jflex.de/manual.html#Grammar):

> ```
> CharClassContent    ::= CharClass | Character |
>                         Character'-'Character | 
>                         MacroUsage | PredefinedClass
> ```
> a character class [...] matches any character in that class. A Character is considered an element of a class if it is listed in the class or if its code lies within a listed character range Character’-’Character or Macro or predefined character class.

lsf37:
I've been looking at it and it will need a refactor of a whole bunch of code. Nothing too hard, just a lot of work. Still planning to do that, but it's definitely going to be more than 1-2 months, given that I won't have much time to actually work on it.

Jezza:
Ah, that's good to hear.

I thought it was just dead.  
There's no rush, I've actually just spent the last hour or two restructuring the lexer so I don't have to manually expand them.  
It makes it nicer in the end, so I guess it worked out in the end.

lsf37:
Excellent, good to hear that you're not blocked on it any more. Still planning to work on it anyway, of course.

wdanilo:
Is there any progress with this issue? It's really bad thing right now. Even allowing macros which point to simple character classes would help design the lex input much much nicer.

lsf37:
Main refactor is done and first rough implementation of the feature is there. Currently in debugging.

regisd:
Let me know when you need a code review.

PR #605, opened 2019-11-24
Make `yychar` long

### LINKED ISSUE #536: Lexer has negative yychar when processing files larger than 2 GiB

regisd:
Fix #536.

When the input is larger than 2GB, the scanner internal state for the number of read characters (aka `zzchar`) is negative. This obviously doesn't make sense.

The test added in #603 reproduces the problem and throws an exception, when the `zzchar` is negative.

The problem is caused by `yychar` being an `int`, hence subject to overflow.

Fix #558 by making `yychar` long. This is similar to #558 by @sarowe with a lighter approach for testing.

Note that some problems remain:
- `yyline` and `yycolumn` are still  integers.
- if the match section is larger than 2GB, then
  - you will likely have an OutOfMemory error
  - even if you have enough heap size, `zzRefill()` will throw a `NegativeArraySizeException` like this
    > java.lang.NegativeArraySizeException
    > 	at jflex.testcase.large_input.LargeInputScanner.zzRefill(LargeInputScanner.java:288)
    > 	at jflex.testcase.large_input.LargeInputScanner.yylex(LargeInputScanner.java:601)
    > 	at jflex.testcase.large_input.LargeInputTest.consumeLargeInput(LargeInputTest.java:22)
    > 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)


regisd:
Thanks for the positive feedback, Gerwin.

My opinion is that:

- It's OK to do API breaking changes, particularly because there was none in the last year (with a slow release cycle, changes can/will be more brutal), and very few in the  last decade.
- There is no need to announce much in advance. Users have to read the changelog and update their code when they upgrade the dep.

That being said, I think 1.8 has accumulated enough changes to be released.

lsf37:
Yes, I meant making it prominent in the release announcement (just because API breaking changes are rare for JFlex), I agree that we don't want to announce it before.

lsf37:
Forgot to say: I agree that 1.8 is getting there. Still labouring under the illusion that I'll manage to pull in the char class macros (#216) into 1.8. Recently made some progress on that, but there is still some way to go.

regisd:
Follow-up comment for those who hit this breaking change. There are two options to fix it:
- either adapt the consuming code and use `long` as well.
- or there is a guarantee that `yychar` is neither larger or equal to 2^31 in which case it's possible to wrap `yychar` with [Math.toExactInt()](https://docs.oracle.com/javase/8/docs/api/java/lang/Math.html#toIntExact-long-)

PR #642, opened 2019-12-01
Bazelify testcase/eofclose

regisd:
Actually rewrite as a simple integration test, no need for golden files.

Migrate both eofclose & eofclose2 as they are about the same directive.


javatests/jflex/testcase/eofclose/EofCloseTest.java
```diff
@@ -0,0 +1,51 @@
...
+/** Test for {@code eofclose} directive. */
+public class EofCloseTest {
+
+  @Test
+  public void eofcloseDirective_true_closesReaderAtEof() throws Exception {
+    Reader reader = readContent();
+    ScannerFactory<EofClose> scannerFactory = ScannerFactory.of(EofClose::new);
+    EofClose scanner = scannerFactory.createForReader(reader);
+    while (scanner.yylex() != EofClose.YYEOF) {}
+    assertThrows(
```
regisd: generally, I think the regression tests should be based on expected _behavior_ rather than the _content_ of `System.out`.
In this case, it's cleaner to the reader throws an exception rather than doing e.printStackTrace() and look for the System.out.

lsf37: I agree, if it can be expressed nicely, it's better to assert behaviour rather than output. The old framework was trying to avoid having to type out each test case in Java. It's about two decades old though, and a lot has happened in Java since.

PR #643, opened 2019-12-01
Fix import of ScannerException

regisd:
Unfortunate merge conflict because #623 was tested long before #601.


regisd:
Since, this is the second time after #619 , I've enable guthub protection

> **Require branches to be up to date before merging**
> This ensures pull requests targeting a matching branch have been tested with the latest code.

PR #644, opened 2019-12-01
Expose `yyatEOF()` in generated scanner API

regisd:
I don't think there is a convenient way to known whether a scanner has finished reading its input.

But it's sounds to me that looping until EOF is a critical use case, which is even implemented in the standalone `main` method as:
```java
while ( !scanner.zzAtEOF ) scanner.yylex();
```
See [Emitter.java:350](https://github.com/jflex-de/jflex/blob/0d6c600861db3cd906c75552b0b0d50ca6f331a7/jflex/src/main/java/jflex/generator/Emitter.java#L350)

This change adds a getter for the `zzAtEOF` field.


regisd:
The benefit is quite clear on [EofminGoldenTest.java](https://github.com/jflex-de/jflex/pull/644/files#diff-e3b733af8d3c88d352daa0d3728548a2)

regisd:
@lsf37 Am I missing something? Can a client of the scanner know whether the scanner has reached EOF?

regisd:
@lsf37 I chose name `zzAtEof()` to be consistent with the field name, but why a `zz` prefix? It could also be `yyAtEof()` or just `isAtEof()`. What do you prefer?

lsf37:
The `zz` prefix means that something is internal, i.e. not part of the API. If we make it a getter method that we want people to use, it should therefore be `yyAtEOF()`. We could also rename the field, but it makes sense to leave that internal, because we might change the mechanism by which we figure out the end of file.

I don't fully remember any more wether relying on `zzAtEOF` is always safe. Will investigate. 

The whole `yy` and `zz` naming scheme is from the `JLex` days (at least `yy` is, and then I think I used `zz` to distinguish). The idea was to avoid conflicts with names that users might want to pick. For the actually internal parts that makes sense, but the API would probably look less strange if it didn't have a prefix. In any case, for new things we should stay consistent.

lsf37:
Ok, after looking through the emitter and skeleton code, `yyAtEOF()` should be fine to export as API. I have vague memories of not doing that initially, because usually the scanner should return a token to indicate that it is done instead of the caller looking at scanner state, but that is really more a matter of taste, so no reason not to export. (Debug and standalone scanners used `zzAtEOF` because it was hard to figure out what the spec uses as EOF token).

lsf37:
We should probably add it to the manual as well. I can do that if you like (happy for you to merge first).

regisd:
> Oops, last comment was for the wrong thread. I'm still investigating, but it looks like it's fine so far. Would only change `zzAtEof()` to `yyAtEOF()`.

Renamed to `yyatEof()` (lower case 'a' to be consistent with the rest) in commit 5f4dd702d53718fe31eccfebe914bc4813472bdd.

regisd:
> We should probably add it to the manual as well. I can do that if you like (happy for you to merge first).

Documentation should be added with code. Done in 44059777c19dac8712663bf78c07968e37f82fbf.

PR #650, opened 2019-12-04
normalise char class order before NFA construction

lsf37:
This is to make generated automata more easily comparable, in particular to make
them more obviously equal when they are isomorphic up to input renaming.

Will hopefully make the upcoming char class refactoring more robust to test.


lsf37:
Codacy doesn't like the explicit `NullPointerException`, but the contract for `compareTo` requires a `NullPointerException` when the argument is `null`, so I'll leave that in.

PR #657, opened 2019-12-05
Modularize build: charset, options

regisd:
* Move Out in jflex.logging
* Move Options in jflex.option
* Create bazel target for jflex.core.unicode
* Move the charclass related code in jflex.core.unicode
* Introduce ILexScan interface to decoupe the unicode classes from the actual LexScan implementation (which is in jflex.core)
* Create jflex.core.OptionUtils for setting the complex options. setDefaults() must be called before the generation is invoked, and this is guarded with the encoding field.
* Reduce bazel target visibility
* Moved DEBUG const in Build class


testsuite/testcases/src/test/cases/java/java-2.input
```diff
@@ -134,7 +134,7 @@ final public class NFA {
   public void addRegExp(int regExpNum) {
 
     if (Options.DEBUG)
-      Out.debug("Adding nfa for regexp "+regExpNum+" :"+Out.NL+regExps.getRegExp(regExpNum));
+      Out.debug("Adding nfa for regexp "+regExpNum+" :"+Env.NL+regExps.getRegExp(regExpNum));
```
lsf37: The search/replace was a bit too global here, this file should remain the same.

jflex/src/main/java/jflex/env/Env.java
```diff
@@ -0,0 +1,9 @@
+package jflex.env;
+
+public class Env {
+
+  private Env() {}
```
lsf37: Codacy wants the constructor after the fields. Probably fair.

regisd:
> Good idea in general, just minor things to fix up (one the input files in the testsuite happens to be java code from some old JFlex version that mentions Out -- that should remain as is)

This is WIP. I want to actually to do a more serious modularization. NL was just a small annoyance. I hope I'll get something.

lsf37:
No worries.

regisd:
Just pushed my latest update, this is basically moving
- Out in logging package
- char-classes related code in unicode package
- core depends on both logging and unicode packages

The problem is still with `Options` which remains in core. Some Out methods depend on Options. I can't have cyclic dependencies.

jflex/src/main/jflex/LexScan.flex
```diff
@@ -13,9 +13,9 @@ import java_cup.runtime.Symbol;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.FileReader;
-import jflex.core.unicode.UnicodeProperties;
+import jflex.core.unicode.IntCharSet;import jflex.core.unicode.UnicodeProperties;
```
lsf37: typo

jflex/src/main/jflex/LexScan.flex
```diff
@@ -13,9 +13,9 @@ import java_cup.runtime.Symbol;
...
 import jflex.l10n.ErrorMessages;
-import jflex.performance.Timer;
+import jflex.logging.Out;import jflex.performance.Timer;
```
lsf37: typo

lsf37:
> Just pushed my latest update, this is basically moving
> 
> * Out in logging package
> * char-classes related code in unicode package
> * core depends on both logging and unicode packages

In principle no problem with that.

> The problem is still with `Options` which remains in core. Some Out methods depend on Options. I can't have cyclic dependencies.

Hm. This is tricky. I don't see how logging should not depend on runtime options. Maybe from the other direction. Maybe a separate module for Options and related?

lsf37:
Looks fine so far. Nitpick: package `jflex.state` should be `jflex.states` to avoid confusion between program state and NFA/DFA states.

lsf37:
Sorry about the conflict, just a comment change.

lsf37:
(merged this to avoid conflicts, hope that was fine)

PR #658, opened 2019-12-07
Use Factory methods in Interval and IntCharSet

regisd:
Follow-up of #656 
See _Effective java_, item 1.

### DIFF
--- a/jflex/src/main/java/jflex/core/IntCharSet.java
+++ b/jflex/src/main/java/jflex/core/IntCharSet.java
@@ -46,6 +46,17 @@ public final class IntCharSet implements Comparable<IntCharSet> {
   }
 
   /**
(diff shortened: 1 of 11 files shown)


jflex/src/main/java/jflex/core/IntCharSet.java
```diff
@@ -185,43 +190,7 @@ public void add(Interval interval) {
    * @param c Character to add.
    */
   public void add(int c) {
```
lsf37: This was a specifically optimised version for a single character with simpler logic for merging intervals. I'd like to keep that. Can you please revert this bit?

lsf37: Actually, don't worry, I have another commit incoming and will put it into that one.

PR #661, opened 2019-12-08
simple invariant checking

lsf37:
Currently for Interval and IntCharSet only, but would make sense for all set and partition implementations in JFlex.

This aim is to replace the horrible old printf debugging with contract and invariant checking over time.

This is separate from unit testing those implementations. The idea is that these assertions are for debugging only and will help find where exactly things have gone wrong.


lsf37:
I've kept the `invariants` methods to type boolean, so that they can be reused in unit tests.

Issue #662, opened 2019-12-08
Property-based unit tests


lsf37:
The current unit testing in JFlex is a bit thin, mostly because the cost/benefit ratio of writing those single test cases is fairly low.

We should use property-based unit tests, at least for the different set implementations in JFlex, as well as partitions and other mathematical objects that have nicely statable properties. These would automate the test object construction, and let us specify the contracts for those objects in a way that gains some actual assurance about implementation correctness.

juni-quickcheck (https://github.com/pholser/junit-quickcheck) looks like it would be usable for the job.

lsf37:
I'll have a look at how much work it would be to set up. @regisd, @sarowe do you have any opinions on this?

regisd:
I discovered this kind of testing when I learned scala, and liked it, but never used it in real life.

I'm a bit uncertain about the randomness of things ;  as in can this make tests flaky?

lsf37:
In theory the randomness could lead to a test that does not always fail, in practice that happens rarely, although it does happen. How often it happens depends on the number of tests it runs for each checked property (default is 100), and how well the input space is covered in that number of tests. So far, if something was wrong, it failed pretty much always, but it's not guaranteed.

When a test fails, it shows the inputs and random seeds for which it failed, so it can be reproduced. It's usually a good idea to then turn that failing instance into a unit test for reliably checking regressions. E.g. for IntCharSet, you'd take the parameters it shows, construct them manually in the old `IntCharSetTest` unit test, but call the `IntCharSetProperties` method that failed with it, i.e. you don't have to duplicate the actual test code. This will unit test will now always fail until the problem is fixed.

lsf37:
It seems to be working fine, and it's at least for me way easier to find good tests to write, so I'll continue working on this a bit and add tests for state sets and char classes.

lsf37:
#687 bring this to reasonably comprehensive tests for `IntCharSet`, `StateSet`, and `CharClasses`. I'll see if it make sense to do something on `RegExp` next, and if that is successful, on NFA and DFA.

PR #663, opened 2019-12-08
Quickcheck

lsf37:
Adds property-based tests for `IntCharSet` (see also #662).

Seems to work really nicely, reasonably fast, fails pretty much immediately on any property misbehaviour. I even discovered a few more assertions that should be there for debugging.


lsf37:
@regisd would you be able to add the corresponding dependency to the bazel setup?

regisd:
> @regisd would you be able to add the corresponding dependency to the bazel setup?

Yes it was trivial to do.

* Added the third_party declaration in commit 1aa0ee7436deabce4d2003c533ea722cbef17dab
  If you want to do the same in the future, I wrote some documentation in https://github.com/jflex-de/jflex/tree/master/third_party#add-a-new-dependency
* Added the test target in commit 6650a43366c5f282f2db4aa4d7f7496b8fbc7d36

jflex/src/test/java/jflex/core/IntCharSetProperties.java
```diff
@@ -0,0 +1,171 @@
...
+
+/**
+ * Property-based tests for {@link IntCharSet}
+ *
+ * @author Gerwin Klein
+ * @version JFlex 1.8.0-SNAPSHOT
+ * @see IntCharSet
+ */
+@RunWith(JUnitQuickcheck.class)
+public class IntCharSetProperties {
```
regisd: This file doesn't contain define properties ; I think we should just name in IntCharSetQuickcheck.

lsf37:
> > @regisd would you be able to add the corresponding dependency to the bazel setup?
> 
> Yes it was trivial to do.
> 
> * Added the third_party declaration in commit [1aa0ee7](https://github.com/jflex-de/jflex/commit/1aa0ee7436deabce4d2003c533ea722cbef17dab)
>   If you want to do the same in the future, I wrote some documentation in https://github.com/jflex-de/jflex/tree/master/third_party#add-a-new-dependency
> * Added the test target in commit [6650a43](https://github.com/jflex-de/jflex/commit/6650a43366c5f282f2db4aa4d7f7496b8fbc7d36)

Thanks for that. Looking at the commit, I got fairly close, but got the reference name `com_pholser_...` in the BUILD file wrong.

jflex/src/test/java/jflex/core/IntCharSetProperties.java
lsf37: Ah, clash of terminology. The things that quickcheck tests are called properties there (because they are supposed to resemble logical properties that you write for all inputs), and they all declare `@Property`. I can see how this is ambiguous wrt to Java properties, though. Happy with `.*QuickCheck`.

PR #673, opened 2019-12-10
Bazel reorg: Generate LexScan and LexParse in directory jflex/core

regisd:
Even if this works, error-prone complains that generated files `LexScan.java` and `LexParse.java` are in directory `bazel-genfiles/jflex` when they should be in a directory ending by `jflex/core` since they are in `package jflex.core;`.
> bazel-out/darwin-fastbuild/bin/jflex/LexParse.java:6: warning: [PackageLocation] Expected package jflex.core to be declared in a directory ending with /jflex/core, instead found /private/var/tmp/_bazel_regis/297b145b0aac08cfb2c9cf2ea0f2624b/execroot/__main__/bazel-out/darwin-fastbuild/bin/jflex
package jflex.core;
             ^
    (see https://errorprone.info/bugpattern/PackageLocation)

Move the generation rules in jflex/src/main/java/jflex/core/BUILD


lsf37:
I like the idea (had a brief look at how to remove those warnings, but didn't continue).

Are the maven plugin changes related? They seem to be the cause of the test failure.

regisd:
> Are the maven plugin changes related? They seem to be the cause of the test failure.

Actually, the jflex-cup-plugin now tries to generate the parser for `jflex/core/BUILD`, so yes, the change is  required to avoid this.

PR #677, opened 2019-12-11
Bazelify testcase/line-cont

regisd:


lsf37:
(see also the new #679)

regisd:
Indeed, the automatic migration is for checking the golden input/output files. But this test didn't have any.

If we care about generation failure (e.g. invalid syntax in the .flex spec), or about the logs of the Jflex generation (like here), then the [`JflexTestRunner` is the right approach](https://github.com/jflex-de/jflex/wiki/Migration-to-Bazel#migrating-a-test-where-the-scanner-generation-fails).
commit 1a966d028a790b59259e3a91f0d6c847256e426b shows how to use it here.

PR #678, opened 2019-12-11
Remove maven test cases which have been migrated to Bazel

regisd:
Unlike Subversion or Mercurial, git doesn't understand when a file is copied.

From now on, for the migration, I think it would be cleaner to move files rather than copying them.


regisd:
Yes, I wanted to ask your opinion on this.

This change was trivial. We can close for now.

PR #682, opened 2019-12-12
do not create char classes for NL characters for atBOL matching

lsf37:
the atBOL matching is done in the runtime engine on raw characters and doesn't go through the `cmap`, so we don't need to create classes for that case.


lsf37:
In some cases this can lead to a coarser partition of the input set, which is good (leads to smaller tables and less memory use).

PR #686, opened 2019-12-14
Remove DFA.old_minimize()

regisd:


regisd:
Right. The other idea I propose is to have a DeprecaedDfa class that extends DFA and offers this method.

lsf37:
It looks like the failing tests are all about the new "." at the end of the "minimised" line. The new "." is right, but if you want to make sure it didn't trigger anything else, it might make sense to leave it out for this PR and make a separate one that updates the test output.

I think we're all looking forward to the days where we don't have to compare string output any more :-)

PR #687, opened 2019-12-16
more quickcheck tests

lsf37:
* more tested properties of IntCharSets
* added CharClasses
* added StateSet
* added counter-example shrinking to a few of the generators
* found a small bug in the tests in StateSet (fixed)
* refactor to remove `IntCharSet.getNext()`, tested for equivalence


regisd:
You are on fire  🔥

PR #689, opened 2019-12-18
Bazelify testcase/ccl-16

regisd:


regisd:
Oops. Already in https://github.com/jflex-de/jflex/tree/54f9702e97cdd799bbe633fbb2c7d21e16c89288/javatests/jflex/testcase/ccl_16

Issue #217, opened 2017-10-28
%include files mtimes are not checked by JFlexTask


idodeclare:
Hello,

A useful enhancement will be to compare also any %include files' lastModified() w.r.t. destFile in the anttask.

Regards.

regisd:
Hello, I see you are using POM. Can you consider using the jflex-maven-plugin? I believe it shouldn't have this issue.

idodeclare:
Thank you — I already have an ant workaround.

This issue is strictly regarding an issue with JFlexTask.

PR #693, opened 2019-12-22
Check %include files' modified times for rebuild too

idodeclare:
Hello,

Please consider for integration this patch for the jflex-maven-plugin to check also last-modified times for `%include` files in a JFlex file when possibly short-circuiting a rebuild. (It had been suggested in #217 that the jflex-maven-plugin supported this already, but I did not find that to be the case.)

Only `%include` files directly referenced by a `lexFile` would be checked by this patch and not any nested `%include` statements.

Thank you.

### DIFF
--- a/jflex-maven-plugin/src/main/java/jflex/maven/plugin/jflex/ClassInfo.java
+++ b/jflex-maven-plugin/src/main/java/jflex/maven/plugin/jflex/ClassInfo.java
@@ -10,6 +10,9 @@ package jflex.maven.plugin.jflex;
 
 import com.google.common.base.Strings;
 import java.io.File;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 import java.util.Objects;
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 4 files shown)


lsf37:
Thanks for your contribution! I’ll give @regisd some time to review, but this looks good from my side. We should probably add nested include files afterwards in a separate commit.

idodeclare:
Thank you for the review, @lsf37. I'm going to push a tweak to more exactly align the plugin's "guessing" regex matching with the LexScan.flex syntax for `%include`:

```
^ {WSP}* "%include" {WSP}+ .*  { includeFile(yytext().trim().substring(9).trim()); }
```

and to fix a typo.

(Would you like me to create an issue for considering nested includes for last-modifed checking?)

lsf37:
Thanks for the tweak, you're right that's closer to the expression in JFlex. 

> (Would you like me to create an issue for considering nested includes for last-modifed checking?)

Yes, please. I was going to do it right after the merge, but better to make sure I don't forget.

Looking through the code again, one question: is there a deeper reason for ignoring the include files in `equals` and `hash`? My instinct would be to add them here as well.

idodeclare:
> Looking through the code again, one question: is there a deeper reason for ignoring the include files in `equals` and `hash`? My instinct would be to add them here as well.

Oh my thinking was that the existing `className` and `packageName` were fundamental to the identity of `ClassInfo` as their natural key — and so part of the hash — but that the new tagalong data of `includedFiles` was not.

lsf37:
> Oh my thinking was that the existing `className` and `packageName` were fundamental to the identity of `ClassInfo` as their natural key — and so part of the hash — but that the new tagalong data of `includedFiles` was not.

Got it. The class is currently only used to determine the output file name, but it makes sense the store the include file data here as well. I might rename it to `SpecInfo` later, but will merge as is for now.

@regisd is probably enjoying the holidays, so I'll merge now and we can address any changes later when he returns.

lsf37:
Thanks again for the contribution!

idodeclare:
My pleasure! Thank you for the collaboration

PR #696, opened 2019-12-29
address more errorprone warnings

lsf37:


jflex/src/main/java/jflex/core/AbstractLexScan.java
```diff
@@ -21,7 +21,9 @@
   int bufferSize = 16384;
 
   File file;
-  private Stack<File> files = new Stack<>();
+
+  @SuppressWarnings("JdkObsolete")
+  private final Stack<File> files = new Stack<>();
```
regisd: I'm pretty sure `ArrayDeque` could offer the same features.

jflex/src/main/java/jflex/gui/GridPanel.java
```diff
@@ -20,16 +20,15 @@
...
+  private final int rows;
 
-  private int hgap;
-  private int vgap;
+  private final int hgap;
+  private final int vgap;
 
-  private Vector<GridPanelConstraint> constraints = new Vector<>();
+  private final Vector<GridPanelConstraint> constraints = new Vector<>();
   private Insets insets = new Insets(0, 0, 0, 0);
```
regisd: Consider using `ArrayList`

lsf37: Yup, should. That's a separate warning, though ;-)

jflex/src/main/java/jflex/core/AbstractLexScan.java
lsf37: Happy in principle to change out Stack for ArrayDeque.

lsf37:
I have another set coming up that addresses both of these.

Issue #199, opened 2016-05-13
%unicode 2.0 lexers throw IOOBE on input with surrogate chars


gregsh:
Default %unicode means 10x times static memory footprint  (see zzUnpackCMap).
Switching to 2.0 leaves Character.codePointAt() generation intact leading to:

java.lang.ArrayIndexOutOfBoundsException: 65536

in advance() method.

sarowe:
Surrogates aren't characters, but rather code units that as a pair form a character above the basic multilingual plane, aka a supplementary character.  There were no actual defined supplementary characters in Unicode until v3.1.

I think what you want is:

1) Don't use memory required to handle supplementary characters.
2) Handle supplementary characters.

But you can't have both; you must choose.

lsf37:
@gregsh The static memory footprint is larger than `%8bit`, yes (should be something like 256x, actually), but it's insignificant compared to the memory sizes on modern machines and most of it will never be in the working set, so it's not slowing down anything. If it's really important to have a smaller footprint, you can select `%8bit` or similar.

An exception would be more worrying. Do you have any input that can produce one?

nicity:
@sarowe As there are no surrogate characters in Unicode 2.0 we expect them to be handled by lexer (generated with %unicode 2.0) as usual characters, similar to the older versions of JFlex.

@lsf37 as to insignificance of  lexer's memory footprint: I observe each lexer to increase its memory footprint for 2M. In IntelliJ IDEA CE (https://github.com/JetBrains/intellij-community, one of the product we are developing) there are 27 lexers thus memory footprint is increased for 54M, for ultimate product memory increase is more than 90M. Given that default heap size of IntelliJ IDEA on 32 bit java is 500M the lexer's footprint increase is really huge for us!

lsf37:
Alright, I can see how 27 lexers would increase the overhead problem. You can always use `%8bit`, but it sounds like you want Unicode support. As @sarowe says, I don't think you can have both without significant performance degradation. You could imagine some kind of page-table-like sparse translation map, but you'd more than double the memory accesses per scanned character in the innermost scanner loop.

I assume the lexers don't all run at the same time? If yes, we could delay the creation of the `ZZ_CMAP` to creation of the lexer object, i.e. call `zzUnpackCMap` in the constructor instead of static class initialisation time. That would mean, you only need the memory for those lexers that are actually used (depending on when you call the constructor).

sarowe:
[...]

PR #697, opened 2020-01-01
memory performance: two-level cmap lookup

lsf37:
This addresses the memory issue #199. With more recent Unicode versions, the
character map took 0x110000 * 4 bytes, i.e. ~4MB. Most of this is never
accessed, but it still increases overall memory consumption esp for multiple
scanners.

This PR:
 * changes the current one-level array to a two-level table structure
 * adds quickcheck tests for the table construction
 * updates runtime engine to use that structure
 * updates skeleton files to remove reference to old ZZ_CMAP
 * removes trailing white space in generated template code

Typical memory consumption for the char map decreases from ~4MB to < 100KB,
for simple scanners (little unicode use > 0xFF) to ~20KB.

Even though this increases the number of operations in the innermost loop,
with a bit of luck performance might actually benefit because of better
cache locality. Benchmarking still to be done.

### DIFF
--- a/javatests/jflex/testcase/include/skeleton.nested
+++ b/javatests/jflex/testcase/include/skeleton.nested
@@ -36,7 +36,7 @@
 
   /** the textposition at the last accepting state */
   private int zzMarkedPos;
-  
+
   /** the current text position in the buffer */
   private int zzCurrentPos;
 
@@ -54,12 +54,12 @@
   private int yychar;
 
   /**
-   * the number of characters from the last newline up to the start of the 
+   * the number of characters from the last newline up to the start of the
    * matched text
    */
   private int yycolumn;
 
-  /** 
+  /**
    * zzAtBOL == true <=> the scanner is currently at the beginning of a line
    */
   private boolean zzAtBOL = true;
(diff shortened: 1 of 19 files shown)


lsf37:
This solution is similar to, but (hopefully) simpler than the one implemented for the IntelliJ plugins in #199, with similar memory improvements.

It does not really address the `ArrayIndexOutOfBounds` problem, but this has already disappeared with @sarowe's work on more recent Unicode versions. For `%unicode 2.0`, you will still get an exception for code points >= `0xFFFF`, but that is as it should be, because that Unicode version doesn't have these characters. The later versions (including the default just `%unicode`) that do have higher code points don't throw the exception. This is similar to `%7bit` and `%8bit` scanners throwing the same exception for higher characters.

There is the question whether this is the best behaviour, i.e. one could throw a different exception or report a more user-friendly error, but that is a separate question.

lsf37:
The trailing whitespace removal is a bit out of place here -- it slipped in because my editor is set to stripping whitespace, which I forgot about. I think it's nicer to have them removed, though, so I left it.

lsf37:
ps: huge kudos to @sarowe for the excellent unicode test suite coverage. This and the new quickcheck setup helped a lot in getting that feature implemented much faster than usual.

lsf37:
Will merge this one now, because I have too many things stacking up, but am still interested in feedback if there is time.

PR #708, opened 2020-01-03
activate ant tests in examples

lsf37:
These should now build again, and cause test failure when they don't.
Also removed the cp to lib, which is now already done in the maven build to
make sure lib is always up-to-date.


lsf37:
And of course ant is behaving differently on travis than on my build machine..

lsf37:
The ivy download seemed a bit dodgy on travis, but worked when restarted. If this turns out to be more than temporary, we should revert, but I'll give it a go for now.

PR #709, opened 2020-01-05
[WIP] examples

lsf37:
updating the examples such that all of them have at least a Makefile, ant, and mvn build, and can run from the release.


regisd:
The problem with the "common" directory is that the examples are less usable as example projects.
I'm wondering if we shouldn't simply remove most of the examples to concentrate on  just a few a couple useful ones.

lsf37:
We can think about reducing them, because there does seem to be some overlap now. The initial idea was that each of them demonstrates something specific or is a textbook example from somewhere.

`common` is just there to factor out boiler plate build code and remove duplication. We can make the examples more self-contained again by replicating those parts. It's a trade-off between maintainability for us and self-containedness of the example for the user. For examples, maybe duplication isn't as bad as it would be elsewhere.

PR #711, opened 2020-01-09
Benchmark

lsf37:
Start work on a benchmark suite for JFlex (addressing #698).

Initial numbers indicate that we do have a CPU performance regression from #697 (two-level cmap) of ~18% in the scanning engine with a scanner that has almost no actions. Needs more investigation, but it might mean that we should offer this as an option in addition to the old one-level cmap. The new setup is actually faster on a single cold-start run on small to medium inputs, probably because the one-off cost of initialising the cmap array is high in the old version.


lsf37:
Looks like I'm violating assumptions of source deployment on travis somewhere. Will be offline for a few days and look at it afterwards.

regisd:
> Looks like I'm violating assumptions of source deployment on travis somewhere. Will be offline for a few days and look at it afterwards.

The compilation of branch aggregated-java-sources (you know, the one that allows to compile without bootstrap lib, e.g. for Gentoo) failed because https://repo.maven.apache.org/maven2/org/openjdk/jmh/jmh-generator-annproces-1.22.jar was not found.

scripts/compile-aggregated-java-sources.sh
```diff
@@ -21,6 +21,8 @@ mvnget() {
 mvnget org/apache/ant/ant/1.7.0/ant-1.7.0.jar
 mvnget com/google/auto/value/auto-value-annotations/1.7/auto-value-annotations-1.7.jar
 mvnget javax/annotation/javax.annotation-api/1.3.2/javax.annotation-api-1.3.2.jar
+mvnget org/openjdk/jmh/jmh-core-1.22.jar
```
regisd: The compile-aggregated-java-sources is used to compile jflex outside with generated code. No tests are executed in this branch. I don't think we want  to run a benchmark in this branch. The Maven project should be sufficient.

lsf37: Good point. We should probably just remove the benchmark directory for that one. I'm about to head off out of phone coverage now. Feel free to adjust or leave to me for later.

lsf37: Now done.

PR #721, opened 2020-02-25
Upgrade Bazel version in Cirrus to 2.1.1

regisd:


regisd:
Need to use the "cirrus" branch to test all tasks.

PR #724, opened 2020-02-26
remove doclint for this release

lsf37:


lsf37:
The generated `LexScan.java` from 1.7.0 contains javadoc that produces HTML errors. This is fixed in 1.8.0, but for deployment we're still using 1.7.0 to build, so we will ignore these errors for this release.

PR #746, opened 2020-03-26
Split gui and anttask out of main binary

regisd:
The goal is to make a minimal binary:
-r-xr-xr-x  1 regisd  wheel  4771349 Mar 27 00:15 bazel-bin/jflex/jflex_bin_deploy.jar
-r-xr-xr-x  1 regisd  wheel  6822739 Mar 27 00:15 bazel-bin/jflex/jflex_gui_bin_deploy.jar


lsf37:
So I understand the background: what's the rationale for making a minimal binary?

If we can make it so that users can still double click a main jar in the distribution to get a GUI and run the same one to get CLI options, it should be fine.

regisd:
OK, the real rationale was that I wanted to remove the dependency on ant.
And then, I realized  that the GUI was also  useless in a maven/bazel environment.

With this approach, you would download the jflex_gui_bin_deploy instead.

regisd:
Right, another way of seeing things: Main is unchanged.

commit 2bc7155 introduces a new target that offers a CLI only.

regisd:
Reverted

PR #748, opened 2020-03-28
Minor: Address a few Error prone warnings

regisd:
* Make fields final
   https://errorprone.info/bugpattern/FieldCanBeFinal


regisd:
Duplicated in #749

PR #758, opened 2020-05-03
initialise charClasses for 7bit scanners 

### LINKED ISSUE #756: 7bit switch doesn't work but 8bit and unicode does
I'm trying to generate a specification for a language using `7bit` switch but JFlex won't generate anything. It always says:

> Reading "vc.flex"
> 0 errors, 0 warnings.
>
> Generation aborted.

When I use `8bit` or `unicode` switch, it works fine.

> Reading "vc.flex"
> Constructing NFA : 276 states in NFA
> Converting NFA to DFA : 
> ...............................................................................
> .......................................................
> 138 states before minimization, 121 states in minimized DFA
> Writing code to "Scanner.java"
> 0 errors, 0 warnings.
> 
> Generation finished successfully.

I'm using JFlex 1.8.1.

lsf37:
Char class initialisation for 7bit scanners got lost in one of the refactors between v1.7.0 and v1.8.0.

This PR adds a test for 7bit scanners specifically, and makes sure the initialisation code is called on all paths.

This should fix #756


lsf37:
(fixed formatting)

Issue #720, opened 2020-02-25
jflex-maven-plugin - Regeneration of previously generated classes changes modification date of the output file


bintom:
When I `mvn package` already packaged module I got a log informing `MyJflexFile.java is up to date`, but I noticed its modification date has changed. This triggers Maven compiler plugin to recompile the module. I suggest to not change the modification date if content does not change.

regisd:
There is a PR to work around this bug in Maven https://github.com/jflex-de/jflex/pull/765

lsf37:
I'll leave this open -- #765 does not fix it, it just provides specific dates for the jar file.

lsf37:
So I tried this and couldn't reproduce it. When I do `mvn generate-sources` or `mvn package` and get an "up to date" output from the plugin, the file modification time of the generated target sources does not change and sources are also not recompiled ("Nothing to compile - all classes are up to date" from the maven-compiler-plugin) for me.

`mvn package` does re-run the test suite and re-package the compiled classes, but it does not recompile for me. 

Looking at the code, it does explicitly check for modification time in the plugin (that's when it generates the "up to date" output) and returns without further action from that path.

It's possible that this was indeed a more general maven bug that has been fixed since.

I'll close this one for now, but if you keep seeing this behaviour, please do feel free to re-open.

PR #765, opened 2020-05-16
Make maven build reproducible

hboutemy:
- Hardcode an initial value for `<project.build.outputTimestamp>`
- Update maven plugins to version 3.2+

see https://maven.apache.org/guides/mini/guide-reproducible-builds.html

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -96,6 +96,7 @@
     <maven.compiler.source>${jflex.jdk.version}</maven.compiler.source>
     <maven.compiler.target>${jflex.jdk.version}</maven.compiler.target>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <project.build.outputTimestamp>1</project.build.outputTimestamp>
   </properties>
 
   <issueManagement>
@@ -271,7 +272,7 @@
         </plugin>
         <plugin>
           <artifactId>maven-release-plugin</artifactId>
-          <version>2.5.3</version>
+          <version>3.0.0-M1</version>
           <configuration>
             <autoVersionSubmodules>true</autoVersionSubmodules>
             <preparationGoals>clean install</preparationGoals>
--- a/pom.xml
+++ b/pom.xml
@@ -96,7 +96,8 @@
     <maven.compiler.source>${jflex.jdk.version}</maven.compiler.source>
     <maven.compiler.target>${jflex.jdk.version}</maven.compiler.target>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <project.build.outputTimestamp>1</project.build.outputTimestamp>
+    <!-- activates reproducible builds, will be replaced by release:prepare -->
+    <project.build.outputTimestamp>2023-01-01T00:00:00Z</project.build.outputTimestamp>
   </properties>
 
   <issueManagement>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -93,6 +93,7 @@
     <maven.compiler.source>${jflex.jdk.version}</maven.compiler.source>
     <maven.compiler.target>${jflex.jdk.version}</maven.compiler.target>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <project.build.outputTimestamp>1</project.build.outputTimestamp>
```
regisd, line 96: I don't know if this is a local time or in a timezone. In any case, an ISO_OFFSET_DATE_TIME format would be easier to read and guaranteed to work with timezone differences such as different machines or DST. Can you please change that.

pom.xml
```diff
@@ -266,15 +267,15 @@
...
         </plugin>
         <plugin>
           <artifactId>maven-plugin-plugin</artifactId>
-          <version>3.5</version>
+          <version>3.5.1</version>
         </plugin>
         <plugin>
           <artifactId>maven-release-plugin</artifactId>
-          <version>2.5.3</version>
+          <version>3.0.0-M1</version>
```
regisd, line 278: Is it also needed?

pom.xml
hboutemy: if you want, you can define a meaningful timestamp with something like `2019-10-02T08:04:00Z`
if you use maven-release-plugin, the value will be updated automatically when doing `release:prepare`
if you don't use maven-release-plugin, it will be up to you to update the value or not

lsf37:
Sorry for the delay, I need to read up a bit on what exactly maven is doing there to understand what we're doing. Reproducible build as such is a good idea, I'm all for it.

pom.xml
hboutemy: yes, it is needed: this version of the Maven Release Plugin is the first version able to update `project.build.outputTimestamp` property value during the release

hboutemy:
@lsf37 don't hesitate to ping me if you need any explanation

notice that Reproducible Central is now live and with many releases: https://github.com/jvm-repo-rebuild/reproducible-central
it would be nice to add next release of jflex to the list of independently verified reproducible builds

pom.xml
regisd: Turns out dependable bot has updated this anyways. Can you just resolve the merge conflict to the latest version?

regisd:
I think using a timestamp is a bad practice, but it seems Maven can't do better. LGTM.

regisd:
As far as I understand, this will fix #720. I've updated the change description.

hboutemy:
> I think using a timestamp is a bad practice, but it seems Maven can't do better. LGTM.

you need something reproducible for files timestamp in zip archives: any better idea welcome :)

pom.xml
hboutemy: done

hboutemy:
> As far as I understand, this will fix #720. I've updated the change description.

no, i don't think so

this PR is about getting the same binary JFlex `.jar` output when rebuilding: it does not change anything to how JFlex works

lsf37:
(rebased and resolved conflicts)

lsf37 (approved):
Thanks again for pinging us to do that, and sorry that it took forever..

PR #766, opened 2020-06-16
bump guava version

lsf37:
I don't think JFlex uses any of the code in [CVE-2018-10237](https://github.com/advisories/GHSA-mvr2-9pj6-7w5j), but it's better to use the same version (26.0-jre) across the whole project anyway.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -122,8 +122,7 @@
       <dependency>
         <groupId>com.google.guava</groupId>
         <artifactId>guava</artifactId>
-        <!-- Release 21.0 requires JDK 1.8 or newer. -->
-        <version>20.0</version>
+        <version>26.0-jre</version>
       </dependency>
       <dependency>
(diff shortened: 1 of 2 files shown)


lsf37:
Travis has trouble finding `ivy`. I'll merge this anyway for now, and look at travis in a separate PR.

pom.xml
```diff
@@ -122,8 +122,7 @@
       <dependency>
         <groupId>com.google.guava</groupId>
         <artifactId>guava</artifactId>
-        <!-- Release 21.0 requires JDK 1.8 or newer. -->
-        <version>20.0</version>
+        <version>26.0-jre</version>
```
regisd, line 125: I'd just remove the version here and let it defined once in the dependencyManagement in the parent POM

lsf37: I thought this was the parent pom?

regisd: You are right

PR #789, opened 2020-10-20
Fix test range in StateSetQuickCheck

regisd:
```
jflex/src/test/java/jflex/state/StateSetQuickcheck.java:33: error: [XorPower] The ^ operator is binary XOR, not a power operator, so '2 ^ 58' will always evaluate to 56.
  public void size2nbits(@InRange(minInt = 1, maxInt = 2 ^ 58) int size) {
                                                         ^
```
(see https://errorprone.info/bugpattern/XorPower)

### DIFF
--- a/jflex/src/test/java/jflex/state/StateSetQuickcheck.java
+++ b/jflex/src/test/java/jflex/state/StateSetQuickcheck.java
@@ -30,7 +30,7 @@ import org.junit.runner.RunWith;
 public class StateSetQuickcheck {
 
   @Property
-  public void size2nbits(@InRange(minInt = 1, maxInt = 2 ^ 58) int size) {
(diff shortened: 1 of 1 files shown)


regisd:
There are other problems. All fixed in #788

lsf37:
Well, that is embarrassing. Too much time spent in languages where `^` is indeed power not xor. Thanks for fixing it.

regisd:
Note that I fixed as a "same behaviour". Wanted to ping you about this because I think you actually wanted to use Integer.MAX_VALUE instead (i.e. 2^32 - 1).

lsf37:
Yes, saw that, thanks. I have a todo item to play with the ranges a bit. I was intending to start with max, but if that is any good depends on whether it increases test times or not. Sometimes it's better to choose smaller ranges to get more density.

I'll have a look.

PR #794, opened 2020-12-13
Allow java comment after yylexthrow

regisd:
Currently the `%lexthrow` keyword can only be followed by identifiers
```
"%yylexthrow"{WSP}+ {QUIL} {WSP}*  {...}
```

This change allows the identifier of the exception to be followed by a comment.
```
"%yylexthrow"{WSP}+ {QUIL} {WSP}* {JavaComment}? {...}
```

However, the emitter must also be adapter to emit one exception per line, otherwise the comment of the first might hide all the following declarations.

Also refactor the `AbstractLexScan.lexThrown` field from `String` to `List<String>`, and let the emitter handle the formatting.

Add an integration test.


regisd:
@lsf37 I think this fixes the described bug.

There might be other ways, such as dropping the comment and only keep the identifier.

Also, there might be other keywords to fix.

What do you think?

regisd:
@lsf37 Can you have a look at this PR? It's not very complex, but it's changing jflex core and I don't want to submit without review.

lsf37:
Thanks for doing this, I can look at it tomorrow and finish it off (if there is anything to be changed, which there probably isn't)

regisd:
My knownledge of the jflex grammar is not great.
Maybe there is a more elegant way, and follow yylexthrow by a lex comment rather than a java  comment.

jflex/src/main/jflex/LexScan.flex
```diff
@@ -277,8 +277,8 @@ DottedVersion =  [1-9][0-9]*(\.[0-9]+){0,2}
   "%initthrow" {WSP}+ {NNL}*  { throw new ScannerException(file,ErrorMessages.QUIL_INITTHROW, yyline); }
   "%eofthrow"  {WSP}+ {QUIL} {WSP}*  { eofThrow = concExc(eofThrow,yytext().substring(10).trim()); }
   "%eofthrow"  {WSP}+ {NNL}*  { throw new ScannerException(file,ErrorMessages.QUIL_EOFTHROW, yyline); }
-  "%yylexthrow"{WSP}+ {QUIL} {WSP}*  { lexThrow = concExc(lexThrow,yytext().substring(12).trim()); }
-  "%throws"    {WSP}+ {QUIL} {WSP}*  { lexThrow = concExc(lexThrow,yytext().substring(8).trim()); }
+  "%yylexthrow"{WSP}+ {QUIL} {WSP}* {JavaComment}? { lexThrow.add(yytext().substring(12).trim()); }
```
lsf37, line 280: `{JavaComment}` is correct here (vs `JFlexComment`) if we include it in the generated code, because it's Java going to be code. We could try to separate the comment from the exception (and not include it in the generated code), but it's probably nice to have there and it'd also be fiddly to extract it.

lsf37 (approved):
This looks good. I can't think of a nicer way to include the comments, at least not without a major refactor (the whole section should probably be more in the parser and less in the lexer, but that's for another time).

Happy for you to merge!

jflex/src/main/jflex/LexScan.flex
regisd: OK. Glad I  didn't miss something obvious.

PR #797, opened 2020-12-16
Fix an NPE when generating dot files (jflex-de#769)

Mivik:


lsf37 (approved):
This also looks good, thank you for the contribution!

PR #801, opened 2020-12-18
Add Github Dependabot config

regisd:
This should result in automatic updates of Maven deps
https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates


lsf37 (approved):
👍

PR #812, opened 2020-12-19
UCD Generator of emoji

regisd:
Make the bazel UCD generator handle Emoji
* Add emoji resources in //third_party
* Add a binary scanner for emoji resource
* Add a disabled test of Unicode 10.
  Currently failing with:
  > 1) emitUnicodeVersionXY_10_0(jflex.ucd_generator.UcdGeneratorIntegrationTest) \
  > org.junit.ComparisonFailure: Content differs on line 1480: \
  > : expected:<    "\000\u0[377]"> but was:<    "\000\u0[1f5]">


regisd:
With this change, I believe Unicode v5-v10 are generated like the unicode-maven-plugin used to.

PR #813, opened 2020-12-20
UCD generator: Copy the Script_Extensions at the end of the parsing

regisd:
Optimization compared to doing it on every script addition.


regisd:
Already submitted.

PR #814, opened 2020-12-20
Bazel manual.pdf use pacakges verbatim and lmodern

regisd:
* Use package verbatim
* Use font lm

### DIFF
--- a/docs/manual_full.tex
+++ b/docs/manual_full.tex
@@ -1,12 +1,15 @@
-\documentclass{article}
+%\documentclass[11pt]{scrartcl} % TODO
+\documentclass{report}
 
 \usepackage[T1]{fontenc}
 
 % \usepackage{a4wide} % deprecated. Replace by geometry.
 \usepackage{geometry}
+\usepackage{verbatim}
 \usepackage{graphicx}
 \usepackage{microtype}
 \usepackage{upquote}
+
 \usepackage{color}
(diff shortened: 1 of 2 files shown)


docs/manual_full.tex
```diff
@@ -28,6 +31,11 @@
    margin=2.54cm
 }
 
+% pandoc packages
+\usepackage{lmodern} % Is it necessary? CM looks the same, doesn't it?
```
lsf37, line 35: `lmodern` is there to make sure it's using scalable fonts, not bitmap fonts. Depending on system setup that may already be the default, this just forces it. (One should think it's not necessary any more in 2020, but ..)

docs/manual_full.tex
```diff
@@ -28,6 +31,11 @@
...
+% \usepackage{fixltx2e} % provides \textsubscript % Not necessary?
+% \usepackage[utf8]{inputenc} % Not necessary?
```
lsf37, line 37: Better to have -- if someone accidentally adds non-ascii characters this will make LaTeX understand them.

docs/manual_full.tex
```diff
@@ -28,6 +31,11 @@
    margin=2.54cm
 }
 
+% pandoc packages
+\usepackage{lmodern} % Is it necessary? CM looks the same, doesn't it?
+% \usepackage{fixltx2e} % provides \textsubscript % Not necessary?
```
lsf37, line 36: Yes, this one can be kicked out these days.

PR #802, opened 2020-12-18
Add bibliography to the Bazel-generated manual

regisd:

PR #815, opened 2020-12-20
Remove Makefile to build PDF manual

regisd:
@lsf37 I think the manual generated by Bazel looks reasonably similar to the current version (the issue until now has been fixed with #802).

Shall we remove the old Makefile to build the doc. The github workflow also seems unnecessary, since Cirrus-CI is very efficient in building the doc as part of `bazel build //...`


lsf37 (changes requested):
Sorry, but I don't think the bazel manual looks close enough, it's missing half the settings from the original, has different header fonts etc, and I do definitely *not* want to retire the Makefile build -- I have no idea how to debug latex issues in bazel when things don't work.

Please don't merge this.

regisd:
Bazel is a super-make (with file hashes rather than filestamps, and a cache, and fetching internet resources, and extensibility, and invocation of tests when needed, etc.) but there is no magic. In this case, the [latex rule](https://github.com/ProdriveTechnologies/bazel-latex) only invokes `lualatex`. You will get exactly the same error message than on the command line.

The only thing is that all resources must be defined, and if the package hasn't been used so far, we need to declare it, like [add space pacakge](https://github.com/ProdriveTechnologies/bazel-latex/commit/79e4c381d75e2006259199bf4a1ce52b7b290cd9#diff-062a9d95847921b6399d4a4184d598bb109edc2954926bc86ab9d5627f2a728d) did.

PR #816, opened 2020-12-20
Replace Unicode data by the bazel-generated version

regisd:
I believe the bazel ucd_generator generates the same content as the jflex-unicode-maven-plugin.

The proof that the :ucd_generator works is that:
1. I used the bazel ucd_generator
2. I ran the a UcdGeneratorIntegrationTest that compares the generated files with the current classes.
3. I then overwrote the existing files.

We should observe that
* all tests incl. the regression tests are still passing
* the git diff should only be cosmetic

This change also updates the documentation.

**Only All versions are supported by ucd_generator.** by there are hacks
* Version 2.0 has has  block=arabicpresentationformsb currently has <\ufe70\ufef[e]> whereas ucd_generator wants to output <\ufe70\ufef[f]> #835
* Version 1.1 Integration couldn't very the order of the intervals because of a bug in the test infrastructure. Please double-check the change in git diff.

**Functional changes**
* The Unicode data is for 3.1.1 rather than 3.1.0


regisd:
Oops, I didn't bother to generate older Unicode versions, but they obviously need to be in the binder.

regisd:
@lsf37 Please have a look! These files were generated with https://github.com/jflex-de/jflex/tree/master/java/jflex/ucd_generator

lsf37:
Hm, this is very hard to review on github. The tests passing doesn't mean that they are equivalent, it only means that the tests don't hit the difference if they are there. Maybe we can look at the class files, if they are the same or at least the code and constant pools are the same, we should be good.

Would it be possible to keep the `if` instead of the `case` for now in `UnicodeProperties.java`? Then we might get the same class files, and if that is the case, we can do that replacement in a separate step.

regisd:
> Hm, this is very hard to review on github.

How can I make that easier? I actually went through manual verification and don't think the code has changed.

I could for

1.  a first step of reformatting existing code with google-java-format 
2. generating and running google-java-format on the new code

> The tests passing doesn't mean that they are equivalent, it only means that the tests don't hit the difference if they are there.

I assume there is relatively good code coverage now, but passing tests obviously doesn't mean perfect equivalence.

> Maybe we can look at the class files, if they are the same or at least the code and constant pools are the same, we should be good.


> Would it be possible to keep the `if` instead of the `case` for now in `UnicodeProperties.java`? Then we might get the same class files, and if that is the case, we can do that replacement in a separate step.

Honestly, the switch case is both nicer to read and easier to produce. And that's actually the only change in syntax I have made, I hope this can be reviewed manually.

lsf37:
> > Hm, this is very hard to review on github.
> 
> How can I make that easier? I actually went through manual verification and don't think the code has changed.
> 
> I could for
> 
> 1. a first step of reformatting existing code with google-java-format
> 2. generating and running google-java-format on the new code

That's a good idea, that should cut down the GitHub diff to just the things that actually changes. Let's try it.


> > Maybe we can look at the class files, if they are the same or at least the code and constant pools are the same, we should be good.
> 
> > Would it be possible to keep the `if` instead of the `case` for now in `UnicodeProperties.java`? Then we might get the same class files, and if that is the case, we can do that replacement in a separate step.
> 
> Honestly, the switch case is both nicer to read and easier to produce. And that's actually the only change in syntax I have made, I hope this can be reviewed manually.

Sure, I'm happy with the `case` replacement in general, I was just proposing to review it separately.  But getting both sources into the same `code-format` output will probably make that unnecessary, so let's try that first.

regisd:
Thanks for challenging me! There was a subtle bug with Unicode 6.0 (#824 )

I'm now very confident the change is safe:

1. I've added a test that verifies the generated file has the same values than the prod version (using Eclipse JDT AST, and a basic eval of the fields. See #821 for more details)
2. I've reformatted the code to look similar to the new generator (#822). As a result, the git diff is much easier to review.

PTAL!

sarowe:
Hi Regis, I'll take a look tonight.

regisd:
Hello @lsf37 and @sarowe. I hope you had a good time during the Christmas holidays. Does this change now look good to be submitted?

lsf37 (approved):
This is good from my side, quite a project! Thanks for doing the additional work to make sure.

Issue #767, opened 2020-06-16
Migrate travis to github workflows?


lsf37:
Now that github workflows are free for public repos, it is probably easier to maintain and easier to interact with if we run the travis CI tasks directly on github.

@regisd and @sarowe, unless you see any concerns with that, I might give this a go when there is a bit of time and put up a test pull request.

regisd:
For Maven, it makes perfect sense.

Bazel, CIrrus-CI supports remote cache, and I think Github doesn't.

lsf37:
Happy to leave Bazel on Cirrus, it seems better integrated into Github that Travis in any case.

lsf37:
#768 contains enough progress for this to replace travis as the required check for the regression test suite, but there are a few things still to be set up:

- [ ] set up build badge in README, remove travis build badge
- [ ] matrix build for multiple java versions (todo)
- [ ] possibly a build in the matrix for MacOS (either todo or decide not to do) 
- [ ] deploy github pages (todo)
- [ ] send test coverage results (needs auth token)

After that, we can
- [ ] remove travis build

lsf37:
https://github.com/jflex-de/jflex/pull/1005 has now moved the Bazel build to GitHub actions as well. The cache is implemented locally with the GH cache action.

lsf37:
Deploying the jflex maven plugin site doesn't make that much sense, because it would deploy the snapshot version, not the release version. We had already removed this from Travis quite a while ago, i.e. this doesn't have much to do with migrating from Travis/Cirrus to GitHub actions. Leaving this on manual for now (it's covered in the release docs).

PR #819, opened 2020-12-20
Remove duplicate work from Travis

regisd:
Most CI is now done with Cirrus or Github. #767
Let's save resources and remove duplicate Tasks from Travis.
We actually only need the maven site and the aggregated-java-sources. Remove
- regresions tests
- building examples
- building doc


lsf37 (approved):
Yes, good idea. I'd ideally like to move the rest also to GitHub CI, but ran out of time before work got crazy again.

PR #820, opened 2020-12-20
Revert "Split gui and anttask out of main binary (#746)"

regisd:
This reverts commit a0fe4ed68296ead2dee8ba6a7c4bbfe48a443b50

Causes failure of the regression tests
```
Test [src/test/cases/unicode-space/UnicodeWhiteSpace1_5_0.test] finished successfully.
2371
Test failed, unexpected output: Difference in line 117254
2372
expected :[line: 2027 col: 33 match: -- --]
2373
but got  :[line: 2027 col: 33 match: --[--]
2374
Test output: line: 1 col: 1 match: --/* The following code was generated by JFlex 1.4.1 on 5/26/08 10:22 AM */--
```


regisd:
@lsf37 I have no idea why splitting the main()  method is causing a failure of the regression test. The refactoring looks completely unrelated to unicode-space ; and examples or anything else works.

Is this test a change detector?

regisd:
And it's not even clear to me what test is failing.

lsf37:
It may be picking up a different option combination or default or something like that. I won't have time to look at it in detail before the weekend, but can probably dig into it a bit then. It'd be nice to get the CLI split going.

PR #828, opened 2020-12-22
Generate Unicode 3.1 and 3.2 with Bazel

regisd:


regisd:
With the bazel ucd_generator for Unicode 3.x, I observe a different output than the one we currently have in git.
[test failure](https://github.com/jflex-de/jflex/pull/828/checks?check_run_id=1594547387)

## Unicode 3.0

The scanner doesn't work. I believe that's were the ArchaicScanner is useful, and I didn't port it yet.

## Unicode 3.1

> * propertyValues: Not true that <[alnum, alphabetic, arabic, armenian, asciihexdigit, etc.
> * contains all elements in <[age=1.1, age=2.0, age=2.1, age=3.0, age=3.1, age=unassigned, alnum, alphabetic, arabic, etc.
> * It is missing <[age=1.1, age=2.0, age=2.1, age=3.0, age=3.1, age=unassigned]

@sarowe  Do you remember if there was something specific with this version?

## Unicode 3.2

> *  propertyValueAliases: Not true that <{ahex=asciihexdigit, alpha=alphabetic, arab=arabic, etc.
> * is equal to <{ahex=asciihexdigit, alpha=alphabetic, arab=arabic, etc.
> * The subject has the following extra entries: {blk=alphabeticpresentationforms=block=alphabeticpresentationforms, blk=arabic=block=arabic, etc.

@sarowe Why does the current `Unicode_3_2.java` not contain the block `blk` value aliases?

The value is used in the interval. [L3928](https://github.com/jflex-de/jflex/blob/df4e42579491489c7075506bc4bca9a7a668f99a/jflex/src/main/java/jflex/core/unicode/data/Unicode_3_2.java#L3928)

And the alias is defined in [PropertyAliases-3.2.0.txt](https://www.unicode.org/Public/3.2-Update/PropertyAliases-3.2.0.txt)
```text
# Non-enumerated Properties
# ================================================
age       ; Age
blk       ; Block
```

Shouldn't we emit the alias in that case?

Or is it expected because the value aliases (like arab=arabic) have not been defined in this Unicode version?

regisd:
Ok, found the reason. The **Age** property shouldn't be emitted for future versions.

PR #859, opened 2021-01-03
Migrate unicode tests

regisd:


regisd:
This was a draft change that became too large. Submitted as #856, #857, #858, #860, #861, #862, #863, #864, #865, #866, #867, #868.

PR #880, opened 2021-01-15
Make quiet logging actually quiet

regisd:
Currently, jflex has options
* `--verbose` which enables `Out.println`
* `--quiet` which negates `--verbose`

Options.verbose and Options.quiet are always the opposite of the other, except in the JflexTestRunner.

This is very confusing and this PR refactors Options and logging:
* Replace the Options.verbose and Options.quiet by Options.logLevel
* To match with current behaviour
  - the log level is WARNING by default
  - `--verbose` increases verbosity to Level.INFO
 * Contrary to current behaviour `--quiet` will reduce verbosity to Level.SEVERE.
   - as needed by the JflexTestRunner
   - as wished by Bazel rule


lsf37:
Close in favour of #984, which syncs the behaviour of `JFlexTestRunner` with the rest. We can still implement log levels later (will need a bit more thinking).

PR #913, opened 2021-04-10
Enable nullness check on jflex/dfa and jflex/state

regisd:
Start nullness check verification on jflex/dfa and jflex/state


regisd:
@lsf37 What do you think about enabling checker framework?

lsf37 (approved):
Yes, I think that's an excellent idea!

PR #953, opened 2022-04-20
Property test improvement

jcoultas:
This pull request increases coverage on property tests by improving generators and updating test cases.  Improvements made to the generators to ensure cased characters are chosen with more frequency.  Tests with caseless options have been added as a boolean with updates to allow caseless to operate correctly.  For State Set updates to ensure resize code is called using the new OffsetGen generator.  Also, property and additional code added to State Set to improve coverage.

I am a part of a research group at the University of Illinois Chicago that works towards novel methods to detect bugs in software projects. This project is identified as being a good candidate for application of our research method to improve coverage using property tests. The above issue outlines the findings with a pull request of the enhanced property test to document the observed behavior.


lsf37:
Very nice to see a PR on test improvements! 

I'm about to head out on vacation, so it'll be a bit before I can look at it in more detail, but I definitely appreciate the effort.

lsf37:
Hey @jcoultas, thanks for adding this, I can see what you're trying to do and I approve of the idea.

I'd make corner case values like 0, 1, MAX a little more likely, maybe up 5% even, but otherwise this looks good.

It does however fail to build at the moment, in particular at
```
jflex/src/test/java/jflex/state/StateSetQuickcheck.java:163: error: cannot find symbol
      @From(OffsetGen.class) int largeOffset) {
```

Are you planning to continue work on this?

jflex/src/test/java/jflex/state/StateSetQuickcheck.java
```diff
@@ -181,10 +195,26 @@ public void addStateAdds(StateSet set, @InRange(minInt = 0, maxInt = 34) int e)
...
+    int offset =
+        1001; // note this effected by InRange, so this needs to be adjusted based on annotations
+    // on set, default is set
+
+    //  if no overflow then offset + e, else overflow so use MAX_VALUE
+    int newValue = (Integer.MAX_VALUE - offset) >= e ? offset + e : Integer.MAX_VALUE;
+    set.addState(newValue);
+    assertThat(set.contains(setPre)).isTrue();
+    assertThat(set.hasElement(newValue)).isTrue();
+    assertThat(setPre.contains(set)).isFalse();
```
lsf37, line 217: This condition is not necessarily true: if the new elements that were added happened to be in the set before, the set will not have changed.

jflex/src/test/java/jflex/state/StateSetQuickcheck.java
```diff
@@ -208,6 +238,12 @@ public void complementElements(StateSet s1, StateSet s2) {
     StateSet comp = s1.complement(s2);
     // only elements of s2 are in the complement
     assertThat(s2.contains(comp)).isTrue();
+
+    // ensure that comp does not contain s1
+    StateSet empty = new StateSet();
+    if (!empty.contains(s1)) { // if s1 is {}, then it will always be contained in comp
```
lsf37, line 244: prefer `s1.containsElements()` instead of comparison with a new empty set

lsf37: I don't remember if I have that property already, but a separate property for `!empty.contains(set)` and `!empty.containsElements()` as well as other algebraic properties of the empty set would be a nice addition in general.

jflex/src/test/java/jflex/state/StateSetQuickcheck.java
```diff
@@ -157,14 +159,26 @@ public void removeRemoves(
...
+    // add larger state value to force resize
+    if ((Integer.MAX_VALUE - largeOffset) <= s.bits.length) { // avoid overrun wrap
+      largeOffset = Integer.MAX_VALUE - s.bits.length;
+    }
+    int largerState =
+        ((s.bits.length + largeOffset) << StateSet.BITS) & Integer.MAX_VALUE; // & resets sign bit
+    s.addState(largerState);
+    assertThat(s).contains(largerState);
+    s.remove(largerState);
+    assertThat(s).isEqualTo(sPre);
```
lsf37, line 181: Adding something that forces resize is nice, but I'm not sure we want a resize in every single test case. I think I'd prefer to do this one in a separate property.

Also, the computation here is very dependent on the implementation of `StateSet` (i.e. a bit too white box). It'd probably be nicer to add something to the interface of StateSet that tells us what state would force a resize or what max state can currently be represented, and then we can pick the test case based on that interface.

jflex/src/test/java/jflex/state/OffsetGen.java
```diff
@@ -0,0 +1,45 @@
...
+
+    // 1% change of getting number 1
+    if (rnd == 2) {
+      return 1;
+    }
+
+    // 1% change of getting Integer.MAX_VALUE
+    if (rnd == 3) {
+      return Integer.MAX_VALUE;
+    }
```
lsf37, line 30: I'd up these 3 cases to 5%.

jflex/src/test/java/jflex/core/unicode/IntCharGen.java
```diff
@@ -0,0 +1,171 @@
...
+
+/**
+ * Generator for random Integer values that ensure to sometimes generate a cased character
+ *
+ * @author Jesse Coultas
+ * @version JFlex 1.8.2
+ */
+public class IntCharGen extends Generator<Integer> {
+  // list of ints that have cased values from Unicode 14.0
+  private static final int[] casedList = {
```
lsf37, line 15: There should be a better way to get at these than explicitly listing them for a specific Unicode version. We already have a bunch of classes that represent Unicode versions and encodings. Can we get that information from there?

jcoultas:
@lsf37 I reviewed the build failure and the below test case is setup with seed values that produce the failure.  I am not sure exactly what the issue is with this.  Is this a bug? Or an issue with the updated test? Could you provide me with some insight?

In CharClassesQuickcheck...

```
  @Property
  public void addString(
      @When(seed = -5022752149301929905L) CharClasses classes,
      @When(seed = -4289856472919628505L) String s,
      @When(seed = 4027864125274679581L) @From(IntCharGen.class) int c,
      @When(seed = 3246250712413252630L) boolean caseless)
      throws UnicodeProperties.UnsupportedUnicodeVersionException {

    assumeTrue(s.indexOf(c) < 0);

    if (caseless) {
      classesInit(classes);
    }

    classes.makeClass(s, caseless);
    assertThat(classes.invariants()).isTrue();

    int cCode = classes.getClassCode(c);
    for (int i = 0; i < s.length(); ) {
      int ch = s.codePointAt(i);
      assertThat(classes.getClassCode(ch)).isNotEqualTo(cCode);
      i += Character.charCount(ch);
    }
  }
```

lsf37:
I think this might have uncovered an actual bug, at least so far it doesn't look like anything to do with the test driver. It'd be great if we could minimise the example further to figure out what is actually going on, i.e. what the trigger is.

I've saved the test logs, so it's fine to keep changing things, we won't lose the specific test case.

jcoultas:
@lsf37 That makes sense.  What do you think would be a good assume/assert for the caseless version of addstring?  It seems that we would need write an assume that would ensure that all characters of s, do not include the class of c.  But, not sure how much we would want to lean on the software under test to build up the assume.

jcoultas:
@lsf37 Just checking in to see if you have any ideas or guidance on the above.

lsf37:
Sorry, no super bright ideas on this one. Basically, for caseless you'd need to check that if the class is equal the code point `c` and the code point in the string are case-equivalent. IIRC we have a bunch of Unicode utility classes for that. Of course those classes are used to compute the char classes in the first place, but you'd at least get that the char class construction as such is good (but missing out on wether the classification as case-equivalent is really correct).

jcoultas:
@lsf37 I've updated one test with an assumeThat which seems to allow caseless to run.  Please review and let me know of any issues that would prevent this PR from being merged.

jcoultas:
@lsf37 I actually was able to get addString to fail even with my new assumeThan.  So, I am going to remove the addString updates so we have passing test cases.

jcoultas:
@lsf37  This should be ready to review again.

lsf37:
This is looking mostly good from my side modulo the still unresolved comments (the ones back from 20 Jun). Nice to see the build passing!

jcoultas:
Hi @lsf37 - I think that I have addressed the above items.  Please review and let me know of any noticed issues.

jcoultas:
Hi @lsf37 - Just want to check in for feedback on this PR.

lsf37:
Sorry, things have been busy lately. I'm happy with the StateSet interface and the new changes. What about the [comment](https://github.com/jflex-de/jflex/pull/953#discussion_r901164466) on  `assertThat(setPre.contains(set)).isFalse();`?

When that is addressed, I think we are ready to merge.

jflex/src/test/java/jflex/state/StateSetQuickcheck.java
jcoultas: @lsf37  This line I think will take care of this:
assumeThat(set.hasElement(newValue), equalTo(false));

lsf37: Indeed it will. Thanks, that resolves the remaining issue.

lsf37:
Looks all god now, this has come together really nicely. Thanks again for your contribution and patience.

PR #987, opened 2022-12-31
refactor char class creation

### LINKED ISSUE #986: char class directives delete existing classes
There is an assumption in the char class implementation that the input char set is known before regular expressions are processed, but this assumption is not enforced by the syntax.

In particular, the char class directives (`%7bit`, `%8bit`, `%unicode`, etc) assume they are called only once and before any character set partitions have been constructed. They reset the partitions to one partition covering the (new) whole input char set.

Example:

```jflex
%%
x = ab
%unicode
%%
{x} {}
```

produces

```
CharClasses:
class 0:
{ [0-1114111] }
...
Miniminal DFA is
State 0:
  with 0 in 1
State 1:
  with 0 in 2
State [FINAL] 2:
```

which is wrong and will match any 2-character sequence.

lsf37:
- add test case for #986
 
- move char class creation to after the entire spec has been parsed, so we definitively know the input char set
- to do that, factor out char class creation from scanner and parser into own pass
- to do that, make pre-defined classes and unicode property classes parts of the syntax tree as opposed to inline unfolded char sets
- to do that, factor out regexp char class normalisation into own pass (separate from macro expansion), so that this normalisation can be done after pre-defined classes are converted to char sets
- add error message for declaring input char set twice
- make CharClasses.init() private (use constructor instead)

Fixes #986


javatests/de/jflex/testcase/ccl_init/BUILD.bazel
```diff
@@ -0,0 +1,56 @@
+# ccl3
+
+# This build file was generated automatically, but won't be re-generated.
+# Feel free to improve.
```
lsf37, line 4: remove

jflex/src/test/java/jflex/core/unicode/CharClassesGen.java
```diff
@@ -51,13 +51,11 @@ public CharClassesGen() {
   /**
    * Constructs smallest possible test value, in this case the char class with only one partition.
    *
-   * @return a CharClasses object with one partition.
+   * @return a CharClasses object wit h one partition.
```
lsf37, line 54:
```suggestion
   * @return a CharClasses object with one partition.
```

jflex/src/main/java/jflex/core/RegExp.java
```diff
@@ -304,7 +309,74 @@ public final RegExp normalise(Macros m) {
...
+        unary = (RegExp1) this;
+        return m.getDefinition((String) unary.content).normaliseMacros(m);
+
+      default:
+        throw new RegExpException(this);
+    }
+  }
+
+  /**
+   * Normalise the regular expression to eliminate macro use (expand them), and compound character
```
lsf37, line 354: Update

jflex/src/main/java/jflex/core/RegExp.java
```diff
@@ -341,22 +413,269 @@ public final RegExp normalise(Macros m) {
...
         return new RegExp1(sym.PRIMCLASS, set);
 
-      case sym.MACROUSE:
+      default:
+        throw new RegExpException(this);
+    }
+  }
+
+  /**
+   * Expand pre-defined character classes into primitive IntCharSet classes).
```
lsf37, line 429:
```suggestion
   * Expand pre-defined character classes into primitive IntCharSet classes.
```

PR #1043, opened 2023-01-21
jflex: use smallest type for unpacked arrays

### LINKED ISSUE #196: Reduce RAM usage of the char->char class map
Reduce RAM usage of the char->char class map by using byte[] rather than char[] when the number of classes is less than 257 - see https://issues.apache.org/jira/browse/LUCENE-6913.

This could be an automatic optimization, no option required.

lsf37:
Use byte or char when value range allows to save memory. Runtime impact should be minimal.

Fixes #196


lsf37:
Unfortunately, this makes the generated lexer a good deal slower:

With `byte` for cmap blocks:

```
JFlexBench.noActionLexer         100        1  avgt    5   0.191 ± 0.001  ms/op
JFlexBench.noActionLexer         100        2  avgt    5   0.190 ± 0.004  ms/op
JFlexBench.noActionLexer        1000        1  avgt    5   1.893 ± 0.020  ms/op
JFlexBench.noActionLexer        1000        2  avgt    5   1.877 ± 0.024  ms/op
JFlexBench.noActionLexer       10000        1  avgt    5  19.602 ± 0.072  ms/op
JFlexBench.noActionLexer       10000        2  avgt    5  18.890 ± 0.473  ms/op
```

vs before on 4bd9811:

```
JFlexBench.noActionLexer         100        1  avgt    5   0.168 ± 0.005  ms/op
JFlexBench.noActionLexer         100        2  avgt    5   0.164 ± 0.003  ms/op
JFlexBench.noActionLexer        1000        1  avgt    5   1.663 ± 0.025  ms/op
JFlexBench.noActionLexer        1000        2  avgt    5   1.640 ± 0.074  ms/op
JFlexBench.noActionLexer       10000        1  avgt    5  16.769 ± 0.520  ms/op
JFlexBench.noActionLexer       10000        2  avgt    5  16.671 ± 0.344  ms/op
```

`char` is about the same as `byte`:

```
JFlexBench.noActionLexer         100        1  avgt    5   0.190 ±  0.001  ms/op
JFlexBench.noActionLexer         100        2  avgt    5   0.185 ±  0.002  ms/op
JFlexBench.noActionLexer        1000        1  avgt    5   1.881 ±  0.021  ms/op
JFlexBench.noActionLexer        1000        2  avgt    5   1.840 ±  0.028  ms/op
JFlexBench.noActionLexer       10000        1  avgt    5  19.051 ±  0.122  ms/op
JFlexBench.noActionLexer       10000        2  avgt    5  18.664 ±  0.070  ms/op
```

Therefore I'll close this PR and not merge. We could think about providing this codegen method as an option if memory is a stronger concern than performance for a specific use case.

PR #1106, opened 2023-09-12
Optimize memory usage

hurricup:
- use more memory effective BitSet instead of boolean array
- don't create a buffer until we really need it


ice1000:
Unfortunately the project is dead :( can we have JetBrains jflex published to maven central?

lsf37:
Not sure which project you mean is dead, but I'm not going to add an Emitter change with just the explanation "optimisation" without any benchmarks, use case or anything else provided.

For small scanners this is going to be slower, so from which size on is it better? Is it actually ever better? How much?

ice1000:
> Not sure which project you mean is dead

Ohh!! I'm happy to see you're still active. I assumed this project is no longer receiving updates due to the amount of time it is being inactive.

lsf37:
Still around, just not always with a lot of time. Sorry I was sounding a bit annoyed, PRs with subtle changes to the runtime engine like this one take an inordinate amount of time to review properly compared to the benefit they might (or might not) bring, so I've put it off for quite a while now.

hurricup:
@lsf37 why is it going to be slower?
This was inspired by memory snapshots analysis from intellij idea, which uses slightly modified version of jflex for its lexers.

1. No need to allocate buffers if they are never used
2. Bitset is more memory effective than boolean array

Not sure what benchmarks would you like to see. I did measured performance as well on some huge samples with warmup and stuff, and difference was neglectable, not remember in which direction. Like a percent or so. 

This change is in the intellij fork already, so just wanted to contribute for the common good. If you don't think it is worth it, feel free to drop this PR, i'm totally ok with that.

PR #1111, opened 2024-12-29
Update generated array style

kluever:
C-style array declarations (e.g. `Foo foos[]`) is discouraged by the JLS. Java style (e.g., `Foo[] foos`) should be preferred.


ice1000:
The issue number is 1111! Cool!

lsf37:
Thanks, that will probably quiet down some linters.

